PR #5, opened 2024-03-10
Reorganize sources & Makefile to prepare for integration with LOC framework.

gapisback:
[Build-CI] Reorg code dirs. Separate l3 src/ from sample tests

This commit lays down the groundwork to allow for an easier upcoming integration with another standalone diagnostic tooling.

hanges done are:

- Reorganize code into src/, include/ & test-use-cases/ subdirs
- Create 3 sample use-case test-programs:
        - single-file-C-program/
        - single-file-Cpp-program/
        - single-file-CC-program/

 All 3 programs are identical; the same test.c but reworked as .c, .cpp and .cc source files.

- Implement Makefile rules to build the sample programs and to exercise resulting binaries. Makefile now provides  the following build-dir structure:

        ./build/{release,debug}/{bin,obj}

      `make help` gives complete usage commands syntax.

- Add ./.github/workflows/build.yml to exercise build-and-test for entire set of sources and test binaries.

### Sample `make help` output

```
You need multiple invocations in this series to build all sources:

Usage:

To build C-sample programs and run unit-tests:
 make clean && CC=gcc LD=g++ make all-c-tests
 make run-c-tests

To build C++-sample programs and run unit-tests:
 make clean-l3 && CC=g++ CXX=g++ LD=g++ make all-cpp-tests
 make run-cpp-tests

To build C++-sample .cc programs and run unit-tests:
 make clean-l3 && CC=g++ CXX=g++ LD=g++ make all-cc-tests
 make run-cc-tests

Environment variables:
 BUILD_MODE={release,debug}
 BUILD_VERBOSE={0,1}
```


Makefile
```diff
@@ -1,9 +1,359 @@
 # #############################################################################
-# Simple Makefile to build and test Lightweight Logging Library (l3) sources
+# L3: Simple Makefile to build and test Lightweight Logging Library sources
```
gapisback, line 2: Massively reworked Makefile.

(We've used this previously at VMware, SplinterDB, and this scheme has worked very well.)
I have cloned this version from LOC's Makefile, which itself was derived from SplinterDB's Makefile.

Makefile
```diff
@@ -1,9 +1,359 @@
...
+# ###################################################################
+TMPDIR              := /tmp
+TEST_DATA_SUFFIX    := test.dat
+
+L3_C_DATA           := $(TMPDIR)/$(L3PACKAGE).c-$(TEST_DATA_SUFFIX)
+L3_C_SMALL_DATA     := $(TMPDIR)/$(L3PACKAGE).c-small-$(TEST_DATA_SUFFIX)
+L3_CPP_DATA         := $(TMPDIR)/$(L3PACKAGE).cpp-$(TEST_DATA_SUFFIX)
+L3_CPP_SMALL_DATA   := $(TMPDIR)/$(L3PACKAGE).cpp-small-$(TEST_DATA_SUFFIX)
+L3_CC_DATA          := $(TMPDIR)/$(L3PACKAGE).cc-$(TEST_DATA_SUFFIX)
+L3_CC_SMALL_DATA    := $(TMPDIR)/$(L3PACKAGE).cc-small-$(TEST_DATA_SUFFIX)
```
gapisback, line 130: This collection of test data names are also chosen with `l3` prefix, so `ls /tmp/l3*` will find them easily.

The sample programs are required to supply these names, which they do now.

All tested in CI.

Makefile
```diff
@@ -1,9 +1,359 @@
...
+TEST_CC_CODE_BINS=$(patsubst %program/, %program, $(dir $(TEST_CPP_CODE_BINS_TMP)) )
+
+# Uncomment this to see symbol, for debugging.
+# $(info $$TEST_CPP_CODE_BINS     = [${TEST_CPP_CODE_BINS} ])
+
+# Build symbol for single C & C++ unit-test binary that we run
+C_UNIT_TEST_BIN     := $(BUILD_ROOT)/$(BUILD_MODE)/bin/test-use-cases/single-file-C-program
+CPP_UNIT_TEST_BIN   := $(BUILD_ROOT)/$(BUILD_MODE)/bin/test-use-cases/single-file-Cpp-program
+CC_UNIT_TEST_BIN    := $(BUILD_ROOT)/$(BUILD_MODE)/bin/test-use-cases/single-file-CC-program
+
```
gapisback, line 183: Each of the 3 generated test program binaries are exercised as part of the test targets.

**NOTE: Heads-up**: For next steps of integration with LOC, I plan to add test programs with 2-files, and possibly with multiple-files. This provides the sample-test scaffolding to better demonstrate how you can unpack LOC value into constituent different file names and line numbers.

Hence the choice of names like `single-file-CC-program` and so on.

Makefile
```diff
@@ -1,9 +1,359 @@
...
+	rm -rf $(OBJDIR)/src/l3.o
+
+# ###################################################################
+# Make build targets begin here
+# ###################################################################
+
+all-c-tests:   $(TEST_C_CODE_BINS)
+all-cpp-tests: $(TEST_CPP_CODE_BINS)
+all-cc-tests:  $(TEST_CC_CODE_BINS)
+
```
gapisback, line 205: Note, unlike normal Makefiles, this can no longer supply a single `all` target which will subsume all these `all*` flavours.

This restriction is because of the inability to mix objects built with gcc and g++ in one link step (which I explained at the top).

As this Makefile is mainly for a how-to demo exercise, I think it's ok to live with these multiple `all*` targets.

include/l3.h
```diff
@@ -55,4 +55,7 @@ void l3_log_simple(const char *msg, const uint64_t arg1, const uint64_t arg2);
  *
  * Like l3_log_simple() but does not take arguments. Only available on x86-64.
  */
+ #ifdef __cplusplus
+extern "C"
+#endif
```
gapisback, line 60: Need this when `l3.c` is linked with C++ binaries using g++. 

Otherwise we will get an unresolved reference to this extern fn.

src/l3.c
```diff
@@ -37,14 +37,16 @@ struct l3_entry
...
 {
     uint64_t idx;
     uint64_t fbase_addr;
     uint64_t pad0;
     uint64_t pad1;
     struct l3_entry slots[L3_MAX_SLOTS];
-} *l3_log;
+} L3_LOG;
+
+L3_LOG *l3_log;
```
gapisback, line 49: Are you ok with this naming change?

I think typedef'ed `L3_LOG` for common structure is a pretty normal convention.

I can imagine that as this tooling grows, this L3_LOG may creep into other files, so it's better we finalize and resolve the naming choice, if you were to find this objectionable.

gregthelaw (approved):
Aditya - this is great! Thank you so much for all this work, it's really a big improvement.

PR #8, opened 2024-03-13
[LOC] Integrate with LOC package for improved L3 diagnostic

gapisback:
This commit integrates with Line-of-Code (LOC) diagnostic tooling, which provides a way to encode the code-location where the L3 logging interface is invoked as a 4-byte int.

See https://github.com/Soft-Where-Inc/LineOfCode, which now becomes a git-submodule of L3 repo.

- Prev commit [ @SHA 8283e76 ] has already dropped-in LOC-field as the `struct l3_entry{}.loc_t loc` field definition. The presence of this field in the structure and all L3 APIs  is protected by `#ifdef L3_LOC_ENABLED` guard.

- In this commit, Makefile is enhanced to invoke the build of L3 sample programs under the L3_LOC_ENABLED.

- Use of LOC requires running a Python generator script that comes as part of the LOC package. Makefile is changed to handle  the naming and generation of source headers & file, specific to each sample program being built.

- Update l3_dump.py to unpack log-entry struct contents. For now,  we just unpack the `loc` ID value as-is. (Future commits will  decode this value to filename / line# pair.)

- Update test.sh and CI build.yml to build-and-test all 3 sample programs with default L3 mode (LOC OFF) and with  `LOC-enabled`. Update `make help` info-messages accordingly.

------

### NOTE: To reviewer:

Request a careful review of this integration with LOC package as it sets the stage for how the two tools will interace going forward. Yes, we can always change stuff in future, but will be nice go converge to an acceptable integration, to ease future changes of the integrated outputs.

### Updated help output showing new `make` interfaces:

```
agurajada-Linux-Vm:[107] $ make help

Usage:
[...]
Environment variables:
 BUILD_MODE={release,debug}
 BUILD_VERBOSE={0,1}
 L3_LOC_ENABLED=1

To build L3-sample programs with LOC-enabled and run unit-tests:
 make clean    && CC=gcc LD=g++ L3_LOC_ENABLED=1 make all-c-tests && make run-c-tests
 make clean-l3 && CC=g++ CXX=g++ LD=g++ L3_LOC_ENABLED=1 make all-cpp-tests && make run-cpp-tests
 make clean-l3 && CC=g++ CXX=g++ LD=g++ L3_LOC_ENABLED=1 make all-cc-tests && make run-cc-tests
```

### Sample output from running one of the C++ binaries with LOC enabled

```
---- Run C++ .cpp unit-tests: ----
./build/release/bin/test-use-cases/single-file-Cpp-program

Exercise in-memory logging performance benchmarking: 300 Mil simple/fast log msgs
300 Mil simple log msgs: 13ns/msg (avg)
300 Mil simple log msgs: 8ns/msg (avg)

./l3_dump.py
Usage: ./l3_dump.py <logfile> <program-binary>

python3 l3_dump.py /tmp/l3.cpp-test.dat ./build/release/bin/test-use-cases/single-file-Cpp-program
tid=127708 '300-Mil Fast l3-log msgs' arg1=0 arg2=0
tid=127708 '300-Mil Fast l3-log msgs' arg1=0 arg2=0
tid=127708 '300-Mil Fast l3-log msgs' arg1=0 arg2=0

python3 l3_dump.py /tmp/l3.cpp-small-test.dat ./build/release/bin/test-use-cases/single-file-Cpp-program
tid=127708 loc=65620 'Simple-log-msg-Args(1,2)' arg1=1 arg2=2
tid=127708 loc=65621 'Potential memory overwrite (addr, size)' arg1=3735927486 arg2=1024
tid=127708 loc=65622 'Invalid buffer handle (addr)' arg1=3203378125 arg2=0
```

**NOTE**: that the Python dump script only prints out the LOC-ID value as-is: `loc=65620`.

In future revs, I will unpack this to file name / line#. (That info exists today but I need to do some LOC-python script rework to make it available to L3 dump py script.)


.github/workflows/build.yml
```diff
@@ -20,6 +20,10 @@ jobs:
         echo " "
         which pylint
 
+        echo " "
+        ls -aFlrt
```
gapisback, line 24: Just to check the contents of the repo, now that we are creating LineOfCode as a sub-module of this repo.

.github/workflows/build.yml
```diff
@@ -40,3 +44,12 @@ jobs:
 
     - name: Test-CC-Samples
       run: BUILD_MODE=${{ matrix.build_type }} make run-cc-tests
+
+    - name: Build-and-Run-C-Samples-with-LOC
+      run: BUILD_MODE=${{ matrix.build_type }} make clean && BUILD_MODE=${{ matrix.build_type }} CC=gcc LD=g++ L3_LOC_ENABLED=1 make run-c-tests
```
gapisback, line 49: Retest the sample programs with `L3_LOC_ENABLED=1` to exercise and test this L3-LOC integration.

Here, we just directly try out `run-c-tests` target, which should subsume the build targets, too.

In a way we are exercising and testing another part of the `Makefile` capability.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "LineOfCode"]
+	path = LineOfCode
+	url = git@github.com:gapisback/LineOfCode.git
```
gapisback, line 3: @gregthelaw -- Please review this carefully. I have not worked with submodules myself before, so am not entirely sure if this specification, alone by itself, is sufficient.

Makefile
```diff
@@ -92,6 +98,15 @@ L3_SRCDIR   := $(SRCDIR)
...
 TESTS       := test-use-cases
 
+# Symbols needed to integrate with LineOfCode (LOC) package
+LOCPACKAGE      := LineOfCode/loc
+LOCGENPY        := $(LOCPACKAGE)/gen_loc_files.py
+
+# Names of files generated by LOC Python generator
+LOC_FILENAMES   := loc_filenames.c
+LOC_HEADER      := loc.h
+LOC_TOKENS      := loc_tokens.h
```
gapisback, line 108: This chunk of definitions is where the LOC-package and its artifacts makes its appearance in L3-universe!!

Makefile
```diff
@@ -171,6 +186,12 @@ SINGLE_FILE_C_PROGRAM_SRCS := $(wildcard $(SINGLE_FILE_C_PROGRAM)/*.c)
 # Add L3 package's library file to list of sources to be compiled.
 SINGLE_FILE_C_PROGRAM_SRCS += $(L3_SRCS)
 
+# Name of LOC-generated source file.
+ifdef L3_LOC_ENABLED
+SINGLE_FILE_C_PROGRAM_GENSRC := $(SINGLE_FILE_C_PROGRAM)/$(LOC_FILENAMES)
+SINGLE_FILE_C_PROGRAM_SRCS   += $(SINGLE_FILE_C_PROGRAM_GENSRC)
+endif
```
gapisback, line 193: All dependency of generated sources and targets will be guarded under `ifdef L3_LOC_ENABLED`.

This way, non-LOC-enabled builds will not see the generator coming into play, and there should be no change in behaviour.

Makefile
```diff
@@ -260,6 +299,7 @@ clean:
 	uname -a
 	$(CC) --version
 	rm -rf $(BUILD_ROOT) $(TMPDIR)/$(L3PACKAGE)*$(TEST_DATA_SUFFIX)
+	rm -rf $(L3_INCDIR)/loc*.h $(shell find $(TESTS) \( -name loc*.h -or -name loc_file*.c \) -print)
```
gapisback, line 302: Additionally, `clean` will now clean up any generated files arising from LOC package's usage.

Makefile
```diff
@@ -286,6 +357,14 @@ all: all-c-tests all-cpp-tests all-cc-tests
 INCLUDE = -I ./$(L3_INCDIR)
 INCLUDE += -I ./$(dir $<)
 
+# use += here, so that extra flags can be provided via the environment
+
+# By default, L3-logging runs on its own. To integrate L3 with LOC machinery,
+# to squirrel away the loc_t ID value, run: L3_LOC_ENABLED=1 make ...
+ifdef L3_LOC_ENABLED
+    CFLAGS += -DL3_LOC_ENABLED
```
gapisback, line 365: This line is the magic that enables LOC-tracking within L3 logging APIs.

Makefile
```diff
@@ -286,6 +357,14 @@ all: all-c-tests all-cpp-tests all-cc-tests
...
+    CFLAGS += -DLOC_FILE_INDEX=LOC_$(subst .,_,$(subst -,_,$(notdir $<)))
```
gapisback, line 366: This line is the Real-Magik! behind LOC machinery, which allows compilation to succeed cleanly w/o any work from the user's side.

l3_dump.py
```diff
@@ -31,11 +31,14 @@
 
 with open(sys.argv[1], 'rb') as file:
     data = file.read(32)
-    idx, fibase, _, _ = struct.unpack('<QQQQ', data)
+    idx, loc, fibase, _, _ = struct.unpack('<iiQQQ', data)
```
gapisback, line 34: Unpack the 32-byte struct as {uint32, uint32, uint64, uint64, uint64}, which will now unpack the newly added `loc` ID value also.

src/l3.c
```diff
@@ -20,13 +20,22 @@
...
 
 #ifdef L3_LOC_ENABLED
 #include "loc.h"
 #endif  // L3_LOC_ENABLED
 
+// Define version of static assertion checking that works for both gcc and g++
+#ifdef __cplusplus
+#define L3_STATIC_ASSERT    static_assert
+#else
+#define L3_STATIC_ASSERT    _Static_assert
```
gapisback, line 36: We need this redefinition as this file is compiled using `gcc` with .c samples, but compiled with `g++` when it's built along with C++ sample programs.

src/l3.c
```diff
@@ -35,12 +44,23 @@ struct l3_entry
     pid_t       tid;
 #ifdef L3_LOC_ENABLED
     loc_t       loc;
+#else
+    uint32_t    loc;
```
gapisback, line 48: To simplify unpacking this field from the Python script, always manifest this field by its name.

This way, the `l3_dump.py` does not have "know" if the dump file was created with LOC ON or OFF. 

If L3_LOC_ENABLED is OFF, the field will guaranteedly be 0, so the dumper will skip reporting that value.

src/l3.c
```diff
@@ -81,7 +101,7 @@ l3_init(const char *path)
         return -1;
     }
 
-    l3_log->idx = 0;
+    memset(l3_log, 0, sizeof(L3_LOG));
```
gapisback, line 104: Safe coding. (Not sure why it was not done before.)

src/l3.c
```diff
@@ -112,6 +133,7 @@ l3__log_simple(uint32_t loc, const char *msg, const uint64_t arg1, const uint64_
 {
     int idx = __sync_fetch_and_add(&l3_log->idx, 1) % L3_MAX_SLOTS;
     l3_log->slots[idx].tid = l3_mytid();
+    l3_log->slots[idx].loc = 0;
```
gapisback, line 136: By default, `loc` will be 0.

The user-passed-in argument `loc` value will be stashed away in this field on L138 below.

gregthelaw:
Hi. I just tried this, but it didn't work for me:
```
gel@onyx:~/src/l3$ make clean    && CC=gcc LD=g++         L3_LOC_ENABLED=1 make all-c-tests && make run-c-tests
uname -a
Linux onyx 6.7.5-060705-generic #202402161836 SMP PREEMPT_DYNAMIC Fri Feb 16 19:10:40 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
cc --version
cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

rm -rf build /tmp/l3*test.dat
rm -rf include/loc*.h 

Invoke LOC-generator triggered by:  test-use-cases/single-file-C-program/loc_filenames.c
make: LineOfCode/loc/gen_loc_files.py: No such file or directory
make: *** [Makefile:326: test-use-cases/single-file-C-program/loc_filenames.c] Error 127
gel@onyx:~/src/l3$ find -name gen_loc_files.py
```
are some files missing?

gregthelaw:
Yeah, I've never used submodules before either.
```
gel@onyx:~/src/l3$ git branch
* gapisback/l3-LOC-support
  main
gel@onyx:~/src/l3$ git submodule update
gel@onyx:~/src/l3$ find -name gen_loc_files.py
gel@onyx:~/src/l3$ 
```

gapisback:
Couple of things, @gregthelaw -- I can't see what your branch looks like, and I can't re-try those steps in my version of this branch (as it is already populated with this Git submodule).

I read thru the online [submodule docs, here](https://git-scm.com/book/en/v2/Git-Tools-Submodules). 

Can you check what this looks like for you?

```
agurajada-Linux-Vm:[1259] $ pwd
/home/agurajada/Projects/undoio-l3

agurajada-Linux-Vm:[1260] $ m .gitmodules
[submodule "LineOfCode"]
	path = LineOfCode
	url = git@github.com:gapisback/LineOfCode.git
```

Does this `.gitmodule` file show the above content?

I think to refresh this module in your cloned branch, you will need to run:

```
$ git submodule init
$ git submodule update
```

Or you could have re-cloned entire `le3` repo: `$ git clone --recurse-submodules git@github.com:undoio/l3.git`

But I suppose you are way past that, now.

Also your find command seems incomplete. You need to do:

```
$ find . -name gen_loc_files.py
./LineOfCode/loc/gen_loc_files.py

```

gregthelaw:
@gapisback that worked! i.e. the
```
git submodule init
git submodule update
```
was necessary. We should add that to instructions because other people downloading will need it.

.github/workflows/build.yml
```diff
@@ -40,3 +44,12 @@ jobs:
...
+
+    - name: Build-and-Run-Cpp-Samples-with-LOC
+      run: BUILD_MODE=${{ matrix.build_type }} make clean-l3 && BUILD_MODE=${{ matrix.build_type }} CC=g++ CXX=g++ LD=g++ L3_LOC_ENABLED=1 make run-cpp-tests
+
+    - name: Build-and-Run-CC-Samples-with-LOC
+      run: BUILD_MODE=${{ matrix.build_type }} make clean-l3 && BUILD_MODE=${{ matrix.build_type }} CC=g++ CXX=g++ LD=g++ L3_LOC_ENABLED=1 make run-cc-tests
```
gregthelaw, line 55: Sorry if this is a dumb question: what's the difference between CC and CPP?

.gitmodules
gregthelaw: I haven't used submodules either so am not in a position to review unfortunately. Seems to work though! :)

src/l3.c
gregthelaw: This is redundant - `mmap()` guarantees that the memory it returns will be zero initialized. But, mmap will also map on demand, so doing this will touch all the pages returned by `mmap()` and depending on how much logging our program does, they might not all be used. So I think this change is not needed. Actually, on that note, we could just remove the `l3_log->idx` line, although maybe that's a bit cute.

Apologies, I should have added a comment explaining the rationale here. Perhaps you could add one?

src/l3.c
gregthelaw: Did you measure whether there's a performance impact here? Maybe it will be zero overhead, but it might add a few ns.

gregthelaw: (It's ok if it does add a few ns btw, it's just that we should have a handle on where the time is doing.)

I profiled `printf()` and if you're redicting to a file it's actually quicker than I was expecting -- a simple `printf()` took about 40s and with a couple of integers in the format it took about 140ns. So L3 is still a lot quicker, but not actually by as much as I was expecting.

gapisback:
@gregthelaw - I will respond to your comments and will make the changes needed.

But, meanwhile, could you please review PR #9 for Docs/ updates? I'd like that to go-in first, so I can also address one of your other remarks that `git submodule` dependency needs to be mentioned in our docs. Thanks!

src/l3.c
gapisback: Done. Re-instated the `l3_log->idx = 0;`, with a comment like so:

```
104     // Technically, this is not needed as mmap() is guaranteed to return
105     // zero-filled pages. We do this just to be clear where the idx begins.
106     l3_log->idx = 0;
```

gregthelaw (approved):

src/l3.c
gapisback: We do need this `loc` field to be initialized to something when LOC is enabled. 

I went with your prior comment that `mmap()` will zero-out memory, so have changed the code to check that `assert(loc ==0)` under `DEBUG` build mode.

The new cod chunk will look like this:

```
138 #ifdef L3_LOC_ENABLED
139     l3_log->slots[idx].loc = (loc_t) loc;
140 #else
141 #ifdef DEBUG
142     assert(l3_log->slots[idx].loc == 0);
143 #endif  // DEBUG
144 #endif
```

.github/workflows/build.yml
gapisback: Well ... I was just trying to be careful to support known possible extensions for C++ file names. I've seen `*.cpp`, and `*.cc`.  So `CPP` refers to files ending with `.cpp` and `CC` refers to files ending with `.cc`.

There is also `.cXX`, but did not support that, yet, till I find a code-base that uses that too.

Makefile
```diff
@@ -286,8 +355,20 @@ all: all-c-tests all-cpp-tests all-cc-tests
 INCLUDE = -I ./$(L3_INCDIR)
 INCLUDE += -I ./$(dir $<)
 
+# use += here, so that extra flags can be provided via the environment
+ifeq "$(BUILD_MODE)" "debug"
+    CFLAGS += -DDEBUG
```
gapisback, line 360: @gregthelaw -- Fyi, so you are alerted by this change I sneaked-in as part of the review cycle.

For debug builds, we will now issue `-DDEBUG` in `CFLAGS`.

That will be checked-for in `l3.c`, to assert under DEBUG mode that `assert(loc == 0);`

This ties-in to the comment you made that we should rely on the fact that `mmap()` returns zero-filled memory regions.

src/l3.c
```diff
@@ -112,9 +134,15 @@ l3__log_simple(uint32_t loc, const char *msg, const uint64_t arg1, const uint64_
 {
     int idx = __sync_fetch_and_add(&l3_log->idx, 1) % L3_MAX_SLOTS;
     l3_log->slots[idx].tid = l3_mytid();
+
 #ifdef L3_LOC_ENABLED
     l3_log->slots[idx].loc = (loc_t) loc;
+#else
+#ifdef DEBUG
+    assert(l3_log->slots[idx].loc == 0);
```
gapisback, line 142: IF `L3_LOC_ENABLED` is false (OFF), we now will no longer set `loc = 0` (to avoid minor ns-overhead).

Instead, under `DEBUG` mode , assert() that this field is supposed to be 0.

The Py dump script relies on this field being 0. Otherwsie, in future-revs, that script will try to unpack the contents of a non-zero `loc` field, which, if it's unitialized, will result in returning bogus code-location information.

gapisback:
@gregthelaw -- I see you've passed this review while I was in the middle of addressing your code comments with delta-diffs.

Did you wish to re-review quickly? If not, let me know with a short msg, so I can check this in.

PR #13, opened 2024-03-28
[Test] Makefile rules to build and run unit-tests. Fix l3_dump.py to deal with eof

gapisback:
This commit mainly fixes logic in the l3_dump.py script to process more than a handful of hard-coded number (=3) of log-entries. 

Fix the logic errors in the script to deal with file EOF situation.

Additionally:

- Introduce dir-struct & Makefile rules for adding unit-tests
- Add tests/unit/l3_dump.py-test.c, to exercise above fix in logic.
- Add new Makefile targets: all-unit-tests run-unit-tests
- Enhance CI build.yml to build-and-run new unit-test.
- The main changes are in `tests/unit/l3_dump.py-test.c` and in `Makefile`.


.github/workflows/build.yml
```diff
@@ -27,6 +27,9 @@ jobs:
     - name: Makefile-help
       run: make help
 
+    - name: Build-and-Run-Unit-tests
+      run: BUILD_MODE=${{ matrix.build_type }} make clean && BUILD_MODE=${{ matrix.build_type }} CC=gcc LD=g++ make run-unit-tests
```
gapisback, line 31: Execute newly added unit-tests in CI jobs, too.

Makefile
```diff
@@ -205,11 +223,11 @@ TEST_C_CODE_BINS := $(SINGLE_FILE_C_PROGRAM_BIN)
 ifeq "$(BUILD_VERBOSE)" "1"
     $(info )
     $(info ---- Debug ----)
-    $(info $$SINGLE_FILE_C_PROGRAM_SRCS = [${SINGLE_FILE_C_PROGRAM_SRCS} ])
-    $(info $$SINGLE_FILE_C_PROGRAM_BIN = [${SINGLE_FILE_C_PROGRAM_BIN} ])
+    $(info $$SINGLE_FILE_C_PROGRAM_SRCS = [ ${SINGLE_FILE_C_PROGRAM_SRCS} ])
+    $(info $$SINGLE_FILE_C_PROGRAM_BIN = [ ${SINGLE_FILE_C_PROGRAM_BIN} ])
```
gapisback, line 227: Minor, unrelated, edits for readability of debug output.

Makefile
```diff
@@ -284,6 +302,41 @@ TEST_CC_CODE_BINS := $(SINGLE_FILE_CC_PROGRAM_BIN)
...
+# ---------------------------------------------------------------------------------
+# This invocation will generate makefile-snippets specifying the dependency of
+# each unit-test binary on its sources and required objects. This way, we do not
+# have to enumerate dependency for each unit-test binary built, such as:
+#
+# $(BINDIR)/$(UNIT_DIR)/l3_dump.py-test: $(OBJDIR)/$(UNITTESTS_DIR)/l3_dump.py-test.o \
+#                                        $(OBJDIR)/$(SRCDIR)/l3.o
+# See https://www.gnu.org/software/make/manual/html_node/Eval-Function.html
+# ---------------------------------------------------------------------------------
+$(foreach unit,$(UNIT_TESTBINS),$(eval $(call unit_test_self_dependency,$(unit))))
```
gapisback, line 339: For now, the invocation of the make-helper function `unit_test_self_dependency()` will generate the required dependency as shown on L335.

In future, if we add more unit-test programs, their dependencies will be automatically generated by this clause. So, there should be nothing more to do in the future in this local area.

Makefile
```diff
@@ -355,7 +408,8 @@ $(SINGLE_FILE_CC_PROGRAM_GENSRC): | $(BINDIR)/$(USE_CASES)/.
 	@$(LOCGENPY) --gen-includes-dir  $(L3_INCDIR) --gen-source-dir $(dir $@) --src-root-dir $(dir $@) --loc-decoder-dir $(BINDIR)/$(USE_CASES) --verbose
 	@echo
 
-all: all-c-tests all-cpp-tests all-cc-tests
+all-unit-tests: $(UNIT_TESTBINS)
+all: all-c-tests all-cpp-tests all-cc-tests all-unit-tests
```
gapisback, line 412: Define `all-unit-tests` targeton L411 and include it in `all`.

gregthelaw (commented):
This looks good, but can I suggest splitting into separate commits, one for the tests and another for the product changes? Keeping it as one PR is fine, but I think it's good practice to have separate commits for separate things.

(I find `git reset -p HEAD^` to be a good way to selectively pull things out of a commit, then I do `git commit --amend` to update the commit, and then commit the extracted stuff as a separate one.)

gapisback:
@gregthelaw : Re: "_but I think it's good practice to have separate commits for separate things._"

I have been trained in one-single-commit which keeps the codeline consistent before / after each commit.

If I have to split this work into two commits, I think the test-case commit should go-in first, to demonstrate that there is a bug in the product. Then, the "fix-commit" should come-in next.

The state of `/main` will be stable before / after this pair-of-commits, but stuff will fail if someone just picks the 1st commit.

I prefer to do it this way, even though there will be instability in the branch before the 1st commit. (Going by test-driven dev methodology.)

Let me see how to tease-apart these changes.

gregthelaw:
I agree.

At Undo the policy is to write a failing test that is marked "expect-fail" and then the commit that fixes the bug also removes the expect-fail tag from the test. That's overkill here - if we ever moved to pytest or whatever we could do that, but for now what you suggest sounds good to me.

Makefile
```diff
@@ -26,23 +26,31 @@
...
+	@echo ' '
+	@echo 'To build all sample programs and run unit-tests:'
+	@echo ' make clean && CC=g++ CXX=g++ LD=g++ make run-unit-tests'
+	@echo ' '
+	@echo 'To build all sample programs and run all tests:'
 	@echo ' make clean && CC=g++ CXX=g++ LD=g++ make all'
 	@echo ' make run-tests'
 	@echo ' '
-	@echo 'To build C-sample programs and run unit-tests:'
+	@echo 'To build C-sample programs and run tests:'
```
gapisback, line 43: Edited these messages to say, simply, `run tests` rather than `run unit-tests` as the latter term is now reserved for the newly added unit-test.

Makefile
```diff
@@ -483,6 +540,13 @@ run-cc-tests: all-cc-tests
 	@echo
 	python3 l3_dump.py $(L3_CC_SMALL_DATA) ./$(CC_UNIT_TEST_BIN)
 
+run-unit-tests: all-unit-tests
+	@echo
+	@echo '---- Run L3 unit-tests: ----'
+	./$(L3_C_UNIT_TEST_BIN)
+	@echo
+	python3 l3_dump.py $(L3_C_UNIT_TEST_DATA) ./$(L3_C_UNIT_TEST_BIN)
```
gapisback, line 548: For now, we simply execute the dump script which will only print 3 log-entries. 

I tried to "extend" this script to add code to trip the error, but found all that of not too much value. For now, just live with an incomplete execution.

src/l3.c
```diff
@@ -50,7 +50,7 @@ struct l3_entry
     const char *msg;
     uint64_t    arg1;
     uint64_t    arg2;
-};
+} L3_ENTRY;
```
gapisback, line 53: Minor change to consistently move to `typedef struct` name with UPPER_CASE struct name convention.

l3_dump.py
```diff
@@ -79,13 +85,29 @@ def exec_binary(cmdargs:list):
...
-        row = file.read(32)
+    # pylint: disable-next=invalid-name
+    nentries = 0
+    # Keep reading chunks of log-entries from file ...
+    while True:
+        row = file.read(L3_ENTRY_SZ)
+        len_row = len(row)
+
+        # Deal with eof
+        if not row or len_row == 0 or len_row < L3_ENTRY_SZ:
```
gapisback, line 102: This check is added a bit defensively.

- `not row` is ok; the `file.read()` is doc'ed to return nothing upon eof.
- `len_row == 0` also seems reasonable to check for.
- Not so so sure about `len_row < L3_ENTRY_SZ`. I found this necessary for some tests that would, otherwise, run off into endless loops or print bad output upon reaching what-seems-to-be like an eof.

In any case, if we do hit the 3rd condition, we are looking at an incomplete log-entry, which will be difficult to parse and unpack (anyway).

l3_dump.py
```diff
@@ -10,6 +10,12 @@
 import subprocess
 import os
 
+# ##############################################################################
+# Constants that tie the unpacking logic to L3's core structure's layout
+# ##############################################################################
+L3_LOG_HEADER_SZ = 32  # bytes; offsetof(L3_LOG, slots)
+L3_ENTRY_SZ = 32       # bytes; sizeof(L3_ENTRY)
```
gapisback, line 17: These constants declare a strong-binding between the layout of the C-struct and what this script expects for log-entry sizes.

There should be a better way to check for this binding, but I could not think of an easy way to do that; i.e. the C file has to export these manifest constants.

I will have to some klunky constants in asserts in the .c file and have this script parse for that.

If you have a really good idea for how-to cross-check these bindings, let me know. Otherwise, I'll sit on it for a while and will come back and improve this in future revs.

l3_dump.py
```diff
@@ -95,3 +117,7 @@ def exec_binary(cmdargs:list):
         elif DECODE_LOC_ID == 1:
             UNPACK_LOC = exec_binary([LOC_DECODER, '--brief', str(loc)])
             print(f"{tid=} {UNPACK_LOC} '{strings[offs]}' {arg1=} {arg2=}")
+
+        nentries += 1
+
+print(f"Unpacked {nentries=} log-entries.")
```
gapisback, line 123: FYI: I plan to grep on this message while building pytests (in future) to cross-check the working of this script with known log-file dumps.

gapisback:
@gregthelaw , I have refactored the changes into 2 commits as we agreed to, in [this comment](https://github.com/undoio/l3/pull/13#issuecomment-2029885221).

Kindly re-review at the earliest, so I can move this further along.

Thanks!

l3_dump.py
gregthelaw: I have a recollection that there is some utility to do this, but don't remember it. Agree that hard-coding is the way to go for now at least -- simple is good.

gregthelaw (approved):
LGTM!

PR #16, opened 2024-03-31
Extend fast_log() assembly code to save-off LOC loc_t uint32_t value.

gapisback:
This commit extends fast-logging interfaces, implemented by assembly code, to:

- Support recording 2 args as part of fast-logging.
- Record the code-location ID, LOC-ID, where l3_fast_log() is invoked.
      This is conditionally enabled if env-var L3_LOC_ENABLED=1 is set.

Add new unit-test case and adjust existing use-case sample-programs to exercise these new interfaces.

Co-authored by Greg Law, who reworked l3.S to support this integration.

-----

## Note to the reviewer

Greg -- I squashed your assembly changes developed separately with mine, and added one minor clean-up commit. I'll fold that in , before pushing.

Meanwhile, there is still one more nagging issue. See [this comment](https://github.com/undoio/l3/pull/16/files#r1555049023) of mine. No rush to review this right now; you can get to it once your schedule permits.

### DIFF
--- a/include/l3.h
+++ b/include/l3.h
@@ -38,7 +38,7 @@
 
 /**
  * \brief Initialise the logging library.
- * 
+ *
  * \param path [opt] A filename to back the map where the log will be stored.
  *
  * \return 0 on success, or -1 on failure with \c errno set to something appropriate.
@@ -69,12 +69,21 @@ int l3_init(const char *path);
  * the names of the log file and the executable.
  */
 #ifdef L3_LOC_ENABLED
-void l3__log_simple(loc_t loc, const char *msg,
+void l3__log_simple(const loc_t loc, const char *msg,
                     const uint64_t arg1, const uint64_t arg2);
 #else
 void l3__log_simple(const uint32_t loc, const char *msg,
                     const uint64_t arg1, const uint64_t arg2);
 #endif  // L3_LOC_ENABLED
+
+/**
+ * \brief Caller-macro to invoke L3 Fast logging.
+ */
+#ifdef L3_LOC_ENABLED
+#define l3_log_fast(msg, arg1, arg2) l3__log_fast(__LOC__, (msg), (arg1), (arg2))
+#else   // L3_LOC_ENABLED
+#define l3_log_fast(msg, arg1, arg2) l3__log_fast(L3_ARG_UNUSED, (msg), (arg1), (arg2))
+#endif  //
 
 /**
  * \brief Log a message in as fast a way as possible.
@@ -84,4 +93,11 @@ void l3__log_simple(const uint32_t loc, const char *msg,
 #ifdef __cplusplus
 extern "C"
 #endif
-void l3_log_fast(const char *msg);
+
+#ifdef L3_LOC_ENABLED
+void l3__log_fast(const loc_t loc, const char *msg,
+                  const uint64_t arg1, const uint64_t arg2);
+#else
+void l3__log_fast(const uint32_t loc, const char *msg,
+                  const uint64_t arg1, const uint64_t arg2);
+#endif  // L3_LOC_ENABLED
--- a/l3.S
+++ b/l3.S
@@ -1,29 +1,34 @@
-.globl l3_log_fast // l3_log_fast(msg)
+.globl l3__log_fast // l3__log_fast(msg)
 .extern l3_log
-.extern mytid
 
-l3_log_fast:
-    mov %fs:0xfffffffffffffffc,%eax
-    test %rax,%rax
-    jnz .gotid
-    mov $186, %rax
-    syscall
-    mov %eax, %fs:0xfffffffffffffffc
+l3__log_fast:
+    mov %fs:0xfffffffffffffffc,%eax // Fetch the TLS-stashed TID into %eax
+    test %rax,%rax          // Did we already stash something?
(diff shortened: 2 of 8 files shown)


l3.S
```diff
@@ -1,8 +1,7 @@
-.globl l3_log_fast // l3_log_fast(msg)
+.globl l3__log_fast // l3__log_fast(msg)
 .extern l3_log
-.extern mytid
 
-l3_log_fast:
+l3__log_fast:
```
gapisback, line 4: @gregthelaw - I need your help to rework this assembly logic to support the extended logging interface I am trying to support, as defined in `l3.h` in this commit:

```
     82 #ifdef L3_LOC_ENABLED
     83 #define l3_log_fast(msg) l3__log_fast(__LOC__, (msg))
     84 #else   // L3_LOC_ENABLED
     85 #define l3_log_fast(msg) l3__log_fast((msg), L3_ARG_UNUSED)
     86 #endif  //

[...]

     97 #ifdef L3_LOC_ENABLED
     98 void l3__log_fast(const char *msg);
     99 #else
    100 void l3__log_fast(const char *msg, uint32_t loc);
    101 #endif  // L3_LOC_ENABLED
```

How do I rework the assembly code to save-off the 4-byte loc value?

Can you implement this change in this dev-branch `gapisback/fast_log-LOC-support` and commit the delta? 

I will rework test cases with that assembly support.

gapisback: Ignore L83 for now. Once the assembly is working, I will change all interfaces to become:

`l3__log_fast(const char *msg, uint32_t loc);`

i.e., the LOC-ID can be the 2nd arg to the call, if that's better for the assembly.

include/l3.h
```diff
@@ -69,13 +69,22 @@ int l3_init(const char *path);
  * the names of the log file and the executable.
  */
 #ifdef L3_LOC_ENABLED
-void l3__log_simple(loc_t loc, const char *msg,
+void l3__log_simple(const loc_t loc, const char *msg,
```
gapisback, line 72: Minor change done after you got the `.S` fast-logging interfaces working. Changed this to `const`.

include/l3.h
```diff
@@ -84,4 +93,11 @@ void l3__log_simple(const uint32_t loc, const char *msg,
 #ifdef __cplusplus
 extern "C"
 #endif
-void l3_log_fast(const char *msg);
+
+#ifdef L3_LOC_ENABLED
+void l3__log_fast(const loc_t loc, const char *msg,
+                  const uint64_t arg1, const uint64_t arg2);
+#else
+void l3__log_fast(const uint32_t loc, const char *msg,
```
gapisback, line 101: Minor change done after you got the `.S` fast-logging interfaces working. Changed this to `const`.

Same on L98, above.

tests/unit/l3_dump.py-test.c
```diff
@@ -38,6 +51,21 @@ main(const int argc, const char **argv)
...
+void test_l3_fast_log(void)
+{
+    const char *log = "/tmp/l3.c-fast-unit-test.dat";
+    int e = l3_init(log);
+    if (e) {
+        abort();
+    }
+    l3_log_fast("Fast-log-msg: Args(1,2)", 1, 2);
+    l3_log_fast("Fast-log-msg: Args(3,4)", 3, 4);
+    l3_log_fast("Fast-log-msg: Args(10,20)", 10, 20);
```
gapisback, line 66: @gregthelaw - There is still something going slightly wrong with fast-logging, even when LOC is not enabled.

Unit-tests, by default, run without LOC. 

In [this test output](https://github.com/undoio/l3/actions/runs/8591118511/job/23539502005?pr=16), for the case of `Build-and-Run-Unit-tests`, the value printed for `arg2` for the call on L64 is incorrect:

```
python3 l3_dump.py /tmp/l3.c-fast-unit-test.dat ./build/release/bin/unit/l3_dump.py-test
tid=3102 'Fast-log-msg: Args(1,2)' arg1=1 arg2=93990550136310
tid=3102 'Fast-log-msg: Args(3,4)' arg1=3 arg2=4
tid=3102 'Fast-log-msg: Args(10,20)' arg1=10 arg2=20
```

I added L66, to see if other values are printed correctly; they are.

Strangely, it seems like the very 1st fast-log call's 2nd argument is always being clobbered.

I don't think it's an issue with the Py script that plucks out these values.

I suspect the assembly is stashing something wrong for the very 1st log-entry.

Can you please take a look at this small-item ... whenever you get a chance?

gapisback: This whole cross-checking needs pytests, and I will build some once this core piece gets integrated. (That's a bit of new test infra that needs to be implemented from scratch.)

gregthelaw: OK, I figured out the stupid assembly bug!

gapisback: Applied this fix and verified that this bug is now fixed.

gapisback:
@gregthelaw -- I pulled in your .S fix for the 1st arg clobbering value fix. 

This change-set is now ready to roll. I've squashed all diffs into one commit.

If you can give this a quick final look, I'd like to integrate this to `/main`, so downstream work can be started.

Thanks!

gregthelaw (approved):

PR #20, opened 2024-04-04
[Py] Optimize unpacking LOC-ID when logging Mils of logs

gapisback:
When L3 logging is invoked with `L3_LOC_ENABLED=1`, the unpacking logic invokes the LOC-decoder binary for each LOC-ID found in the logs. For sample programs that log 300 Mil logs, from the same source-code line, this causes a noticeable slow-down in the performance of the Py-dump script.

This commit implements a local optimization to cache the loc-ID from each log-entry, and to skip the LOC-unpacking logic if the `next loc-ID == prev-loc-ID`.

For the 300 Mil logging done by `use-cases/single-file-C-program`, on a Linux-VM, this results in a drop in run-time from ~21s (before this fix) to ~0.2s (with this fix). The run-time is closer to the run-time when dumping the log-entries without L3_LOC_ENABLED, ~0.1624s;  where the loc-ID is printed as-found, without any unpacking.

## Manual testing results

On my Linux-VM, a simple test was done comparing performance of this script while dumping 300 Mil simple-logging log entries generated by the `use-cases/single-file-C-program`:

The commands used to compare dumping 300 Mil log-entries in `/main` v/s this branch are:

```
$ time python3 l3_dump.py /tmp/l3.c-test.dat ./build/release/bin/use-cases/single-file-C-program

$ time L3_LOC_ENABLED=1 python3 l3_dump.py /tmp/l3.c-test.dat ./build/release/bin/use-cases/single-file-C-program
```

| Test                              |      main       |        Fix         |    Remark
|-----------------------|-------------|-------------|------------------------------------------------------|
| L3_LOC_ENABLED=0 | 0m 0.172s    | 0m 0.164s   | Comparable                                                                    |
| L3_LOC_ENABLED=1 | 0m 21.113s   | 0m 0.209s   | Run-time drops dramatically to sub-seconds |

### DIFF
--- a/l3_dump.py
+++ b/l3_dump.py
@@ -91,8 +91,9 @@ with open(sys.argv[1], 'rb') as file:
     idx, loc, fibase, _, _ = struct.unpack('<iiQQQ', data)
     # print(f"{idx=} {fibase=:x}")
 
-    # pylint: disable-next=invalid-name
+    # pylint: disable=invalid-name
     nentries = 0
+    loc_prev = 0
     # Keep reading chunks of log-entries from file ...
     while True:
         row = file.read(L3_ENTRY_SZ)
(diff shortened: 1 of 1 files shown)


gapisback:
@gregthelaw - Minor optimization fix in `l3_dump.py` -- to speed up dev-experience.

You can get to this whenever you have some precious spare cycles.

gregthelaw (approved):
Looks good. Might actually be worth going further and keeping a Python dictionary of known LOCs. It's unlikely that there will be _that_ many different LOC's. But this represents an improvement, so no need to delay getting this in.

gapisback:
Yes, @gregthelaw - I do agree. That thought did occur to me, to build a Python dictionary, but I figured, let's get this simple optimization in first, and then rework the logic depending on how much 'LOC-workload' we generate through more testing.

I prefer to build the test workload to create the situation of large #s of LOCs before building additional optimization. (Test-driven development.)

PR #24, opened 2024-04-15
L3 loc fast

gregthelaw:
Fix daft assembly bug in `l3__log_fast()`


gregthelaw:
@gapisback I'm rather confused with the GH branches here. I pushed the assembly fix as the final commit in this branch but I don't think it's quite right. Can you let me know the intention with how I should branch / create PR's here?

gapisback:
Thanks, @gregthelaw -- for the final fix. I pulled-in [the last commit](https://github.com/undoio/l3/pull/24/commits/1cedd54175d33cd1881c788cb6f83c61a5e7dd8d) to the dev-branch for PR #16 . And have verified that the arg values are coming out right with fast-logging enabled.

We can now close this PR as all your assembly rework will be integrated as part of PR #16.

PR #27, opened 2024-04-20
[MacOSX] Port build infrastructure to run L3/LOC on Mac/OSX

gapisback:
This commit adds a semi-minor overhaul of the build-and-test infrastructure to get the L3/LOC package, along with all newly added Pytest tests, running cleanly on Mac/OSX.

- Tested on macOS Monterey v12.1, and on macos-latest run in CI.

- On MacOSX, gcc is really clang, so with this commit, indirectly,  we also gain full support for clang. We need the `llvm-readelf` binary, which is a LLVM-provided gcc-equivalent binary to readelf.

- **build.yml**: New dependency on llvm-readelf binary that will be installed as part of LLVM install.

  - Add new rules to build-and-test on MacOSX
  - Install required software
  - Main change is to update $PATH to include the path for
    LLVM's llvm-readelf binary

- **Makefile**: We don't, yet, support fast-logging which needs assembly support. So, l3.S will only be included on Linux builds. 
  - Conditionalize build-rules for Linux/Darwin, appropriately.

- **l3.h, l3.c**:
  - Adjust some code to avoid compilation failures seen on Mac
  - On Mac, l3_log_fast() is aliased to l3_log_simple(), just so
    that we keep all programs working unchanged.

- **l3_dump.py**: Major rework here, to deal with llvm-readelf
  (The changes done under 5d75e17 have greatly simplified injecting Mac/OSX support for parsing various readelf outputs.)

  - On Mac, the command we need to use:

    `$ llvm-readelf -p __cstring  ./build/release/bin/unit/l3_dump.py-test`

  - Added `mac_get__cstring_offset()`, to parse the outputs from `llvm-readelf` binary, to reconstruct address / offsets etc.

  - NOTE: This work is really done as part of `l3_init()` for Linux.
  
     But I could not figure the equivalent Mach-O interfaces to get the required offset. Hence, this workaround solution,
     which is still acceptable as it relies on supported `llvm-readelf` binary.

------

### Note to the reviewer:

@gregthelaw - It will be good if you can clone this branch to your local Mac, and let me know if things work.

You will need to install the following s/w components:

```
 brew install python
 brew install pylint
 # To install llvm-readelf
 brew install llvm

pip install pytest
```

I'm not sure about the exact steps to run on your dev-box. Let me know what you find out, and I'll update the docs in a follow-on PR.

This change-set is layered on top of another PR #30. So for this Mac/OSX support, you only need to focus on the 2nd commit in this PR; i.e., https://github.com/undoio/l3/pull/27/commits/35c614635f7e7b59308a7573b25de7171b858d0b

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,21 +1,37 @@
-name: Build and test
+name: L3-Build and test
 on: [pull_request]
 
+#! -----------------------------------------------------------------------------
+# Ref: For running jobs on Linux & Mac/OSX with one workflow.
+# https://stackoverflow.com/questions/57946173/github-actions-run-step-on-specific-os
+#! -----------------------------------------------------------------------------
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: [ubuntu-latest, macos-latest]
         build_type: [release, debug]
+
     steps:
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
         submodules: recursive
+
     - name: Install dependencies
       run: |
-        sudo apt-get update -y
-        sudo apt-get install -y pylint pip
+        if [ "$RUNNER_OS" == "Linux" ]; then
+            sudo apt-get update -y
+            sudo apt-get install -y pylint pip
+
+        elif [ "$RUNNER_OS" == "macOS" ]; then
+            brew install python
+            brew install pylint
+
+            # To install llvm-readelf
+            brew install llvm
+        fi
 
         pip install pytest
 
@@ -29,6 +45,25 @@ jobs:
         echo " "
 
     #! -------------------------------------------------------------------------
+    - name: Update PATH
+      run: |
+        # This is the $PATH reported by `brew install` for the new binaries
+        if [ "$RUNNER_OS" == "macOS" ]; then
+          echo "/opt/homebrew/opt/llvm/bin" >> $GITHUB_PATH
+        fi
+
+    - name: Verify PATH and required readelf binary
+      run: |
+            echo "PATH is $PATH"
+
+            # Verify that this required binary is found in $PATH
+            if [ "$RUNNER_OS" == "Linux" ]; then
+                which readelf
+            elif [ "$RUNNER_OS" == "macOS" ]; then
+                which llvm-readelf
+            fi
+
+    #! -------------------------------------------------------------------------
     - name: test-test-sh-usages
       run: |
         ./test.sh --help
@@ -38,11 +73,16 @@ jobs:
     - name: test-code-formatting
       run: |
         ./test.sh test-pylint-check
-        shellcheck ./test.sh
+
+        # Skip shellcheck on MacOSX. It's throwing up some weird errors!
+        if [ "$RUNNER_OS" == "Linux" ]; then
+            shellcheck ./test.sh
+        fi
 
     #! -------------------------------------------------------------------------
     - name: test-pytests
-      run: ./test.sh test-pytests
+      run: |
+        ./test.sh test-pytests
 
     #! -------------------------------------------------------------------------
     - name: test-make-help
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,10 @@ else
    $(error Unknown BUILD_VERBOSE mode "$(BUILD_VERBOSE)".  Valid values are "0" or "1". Default is "0")
 endif
 
+# Identify the OS we are running on.
+UNAME_S := $(shell uname -s)
+UNAME_P := $(shell uname -p)
+
 # Compilers to use
 CC  ?= gcc
 CXX ?= g++
@@ -190,7 +194,14 @@ BINDIR = $(BUILD_PATH)/bin
 # ###################################################################
 
 L3_SRC      := $(L3_SRCDIR)/l3.c
-L3_ASSEMBLY := l3.S
+
+# Currently, fast-logging assembly is only supported on x86 64-bit Linux.
+ifeq ($(UNAME_S),Linux)
+    ifeq ($(UNAME_P), x86_64)
+        L3_ASSEMBLY := l3.S
+    endif
+endif
+
 L3_SRCS     := $(L3_SRC) $(L3_ASSEMBLY)
 L3_OBJS     := $(L3_SRCS:%.c=$(OBJDIR)/%.o)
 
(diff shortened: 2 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -28,6 +44,25 @@ jobs:
         ls -aFlrt
         echo " "
 
+    #! -------------------------------------------------------------------------
+    - name: Update PATH
+      run: |
+        # This is the $PATH reported by `brew install` for the new binaries
+        if [ "$RUNNER_OS" == "macOS" ]; then
+          echo "/opt/homebrew/opt/llvm/bin" >> $GITHUB_PATH
```
gapisback, line 52: We need to do this up-front as the rest of the tests that fire-up `l3_dump.py` will need to find and run `llvm-readelf`.

Makefile
```diff
@@ -190,7 +194,14 @@ BINDIR = $(BUILD_PATH)/bin
 # ###################################################################
 
 L3_SRC      := $(L3_SRCDIR)/l3.c
-L3_ASSEMBLY := l3.S
+
+# Currently, fast-logging assembly is only supported on x86 64-bit Linux.
+ifeq ($(UNAME_S),Linux)
+    ifeq ($(UNAME_P), x86_64)
+        L3_ASSEMBLY := l3.S
```
gapisback, line 201: @gregthelaw - How hard is it to support assembly for MacOSX? 

I heard there are two diff chips Mac runs on: Intel and something else. Are you comfortable doing this work?

If so, I'd like to log an issue requesting this support, so it can be scheduled sometime in your busy schedule.

Makefile
```diff
@@ -405,14 +418,22 @@ endef
...
 # See https://www.gnu.org/software/make/manual/html_node/Eval-Function.html
 # ---------------------------------------------------------------------------------
+ifeq ($(UNAME_S),Linux)
+
 $(foreach unit,$(UNIT_TESTBINS),$(eval $(call unit_test_self_dependency,$(unit))))
 
+else
+
+$(BINDIR)/$(UNIT_DIR)/l3_dump.py-test: $(OBJDIR)/$(UNITTESTS_DIR)/l3_dump.py-test.o \
+                                        $(OBJDIR)/$(SRCDIR)/l3.o
```
gapisback, line 434: See the comments on L422. I couldn't figure out why this inline Makefile-function is not working on Mac.

It's a minor wart; we can live with for now.

Makefile
```diff
@@ -528,6 +549,8 @@ endif
 
 CFLAGS += -D_GNU_SOURCE -ggdb3 -Wall -Wfatal-errors -Werror
 
+CPPFLAGS += --std=c++11
```
gapisback, line 552: See - here is a soft-benefit, if you will, that this entire tooling can work on even on an older rev-level of C++.

This is unlike the `source_location{}` class stuff that needs C++20.

include/l3.h
```diff
@@ -76,14 +88,27 @@ void l3__log_simple(const uint32_t loc, const char *msg,
...
+#ifdef __cplusplus
+}
+#endif
+
 /**
  * \brief Caller-macro to invoke L3 Fast logging.
  */
+#if __APPLE__
+
+#define l3_log_fast(msg, arg1, arg2) l3_log_simple(msg, arg1, arg2)
```
gapisback, line 100: FYI: For now -- Fast logging on Mac is falling-back to slow-logging interface.

.github/workflows/build.yml
```diff
@@ -38,11 +73,16 @@ jobs:
...
+        # Skip shellcheck on MacOSX. It's throwing up some weird errors!
+        if [ "$RUNNER_OS" == "Linux" ]; then
+            shellcheck ./test.sh
+        fi
 
     #! -------------------------------------------------------------------------
     - name: test-pytests
-      run: ./test.sh test-pytests
+      run: |
+        ./test.sh test-pytests
```
gregthelaw, line 85: I'm not familiar with these files, but just checking: is this deliberate?

Makefile
gregthelaw: Newer Macs have ARM CPUs. It's not fundamentally difficult to support, but I'm not sure there's a huge benefit to doing so. The fast logging was just to get a sense of what a hand-crafted assembly version would perform like. It's not that much faster than the C version.

More troublingly, I've realised there is a problem with the assembly code we have -- if linked against programs that are also using TLS, it will probably conflict. I need to figure out how to craft the assembly in the right way for TLS accesses. This probably isn't too hard, but does need doing.

gregthelaw: Meanwhile, I suggest we live without the fast version on architectures other than x86-64.

include/l3.h
gregthelaw: Yeah, I think this is fine.

gregthelaw: Same should go for 32-bit x86 and any other architectures.

src/l3.c
```diff
@@ -36,6 +44,14 @@
 #define L3_STATIC_ASSERT    _Static_assert
 #endif
 
+#if __APPLE__
+#define L3_THREAD_LOCAL
```
gregthelaw, line 48: I don't think this will work. It'll mean the cached thread id is not thread local, so every thread will report the thread-id of the first thread to call one of our logging functions.

Given that the main idea behind this was a logging library suitable for tracking down race conditions, this is bad.

I'm surprised it's necessary - it's standard C11 so I'd have thought it would be supported on Mac. https://en.cppreference.com/w/c/thread/thread_local

src/l3.c
```diff
@@ -36,6 +44,14 @@
...
+#define L3_GET_TID()  0
```
gregthelaw, line 49: So we'll always report a tid of 0 on Mac? That seems unfortunate (see above).

gregthelaw (changes requested):
I think we should figure out the thread id stuff prior to merge.

.github/workflows/build.yml
gapisback: Yes. This is a test-method to drive all python tests. As a user you don't need to bother yourself with `cd` to some other dir. You can just stay at top-level L3 root-dir and do:

`$ ./test.sh test-pytests`

src/l3.c
gapisback: Yes, you are right. 

Early on when I started this port work, I was a bit lazy trying to figure out the OSX-specific variation to get thread-ID, so just slammed-in something to get things compiling.

I have figured out what to do here, and have also enhanced the tests to verify that the generate thread-ID is != 0.

```
333 def verify_output_lists(nentries:int,
334                         exp_nentries:int,
335                         tid_list:list,
[...]
356     # As in the unit-test a single thread does slow-logging all log-entries
357     # must have the same decoded TID value. And, make sure that L3-logging
358     # code has generated a valid thread-ID, which should be non-zero.
359     assert tid_list[0] != 0
360     for tid in tid_list:
361         assert tid == tid_list[0]
```

Have cross-checked that the tests fail without the change to correct thread-ID.

src/l3.c
gapisback: Have fixed it, like so:

```
+#define L3_THREAD_LOCAL __thread
+#define L3_GET_TID()    pthread_mach_thread_np(pthread_self())
```

gapisback:
@gregthelaw -- I've fixed the thread-local thread-ID issue on Mac/OSX and have confirmed thru new test verification that it's working.

CI-jobs are now passing on both platforms.

As we agreed, I am merging this in even though you have an outstanding review to pass. This will allow me to merge the next PR #31 which is waiting on the runway. You may want to look at the fixes applied in `l3.c` when you get a chance.

PR #29, opened 2024-04-25
[Test] Refactor use-cases sample programs into functions.

gapisback:
This commit refactors inline code in main() of 3 use-cases sample programs into individual functions:

- test_perf_slow_logging() and test_perf_fast_logging()

No other logic has changed. This refactoring is done to enable writing pytests, in future, to verify the outputs of logging.


gapisback:
@gregthelaw - I have a review to you, mainly as an fyi. Do take a look at it if you have the time, but this one is not critical to review.

It's just a code refactoring change I am doing, and will push this now that tests have passed.

PR #30, opened 2024-04-25
[Pytest] Add pytests to verify decoding by l3_dump.py, including LOC-decoding.

gapisback:
This commit adds extensive Python pytests to verify that the output decoded by the Python L3-dump script, `l3_dump.py`, is correct and as expected based on what was logged in source files. 

We also add test-cases that will invoke the unit-test and use-case sample programs with `L3_LOC_ENABLED=1`, which
will trigger default LOC-encoding.

Python test machinery is provided to either validate expected outputs hard-coded in the test-case. Or, validate against
expected outputs generated by running `grep` on the source-files to look for L3-logging APIs, and re-constructing the expected decoded LOC-entry (as string).

In order to facilitate verifying L3-dump outputs using pytests, we do slight refactoring of main() in l3_dump.py into a
separate do_main() method that returns actual-output lists used by the pytest cases to verify against expected-lists.

Sample use-case programs are mildly altered to receive a `--unit-tests` argument, to only run the part of the program that generates a small # of slow and fast logging entries. (The logging done for perf-throughput measurements is skipped under this arg, to keep pytests running fast.)


l3_dump.py
```diff
@@ -227,8 +226,26 @@ def main():
     """
     Shell to call do_main() with command-line arguments.
     """
-    check_usage()
-    (program_bin, decode_loc_id, loc_decoder_bin) = set_decode_loc_id()
+    do_main(sys.argv[1:])
+
+###############################################################################
+def do_main(args:list, return_logentry_lists:bool = False) -> int:
```
gapisback, line 232: For end-user invocation, `return_logentry_lists` will be at its default, i.e. False, so no output-lists decoding will be done.

tests/pytests/l3_dump_test.py
```diff
@@ -0,0 +1,545 @@
...
+                                    tid_list, loc_list,
+                                    msg_list,
+                                    arg1_list, arg2_list)
+    assert verify_rv is True
+
+    # Unit-tests are currently not enabled to run with LOC_ENABLED env-var
+    assert verify_loc_field_is_empty(loc_list) is True
+
+# #############################################################################
+def test_c_test_dump_log_entries():
```
gapisback, line 149: The structure of most test cases here onwards is similar.

- We do some initial boilerplate stuff.
- Build the program
- Execute the resulting binary, generating the L3-log-dump file
- Call the verification method which will do the required cross-checking, for default case (w/ no LOC encoding) and with default LOC-encoding

tests/pytests/l3_dump_test.py
```diff
@@ -0,0 +1,545 @@
...
+                  stdout=sp.PIPE,
+                  stderr=sp.PIPE, text=True) as section:
+        _stdout, _stderr = section.communicate()
+
+    print('\nExec:', cmdargs)
+    print(_stdout)
+    return True
+
+# #############################################################################
+def build_exp_loc_msg_list(prog_dir:str, main_file:str) -> (list, list):
```
gapisback, line 483: This is a **KEY** test-support routine.

We make testing reliable by `grep`-ping actual L3-logging APIs from the source-code, so that the line numbers are automatically, always, correct.

The decoded LOC-ID string format is re-constructed by this method.

use-cases/single-file-C-program/test-main.c
```diff
@@ -33,33 +35,41 @@ timespec_to_ns(struct timespec *ts)
 }
 
 /**
- * The sequence of l3_log_fast() and l3_log_simple() is flipped in this version
- * as opposed to the .cpp and .cc versions of this test exerciser program.
+ * The sequence of l3_logging-APIs is flipped in this version of the sample
+ * program, as opposed to the .cpp and .cc versions of this exerciser program.
  * This is done so that we can see the different 'arg1' values stashed away
- * by the l3_log_simple() call, when unpacked by the l3_dump.py script.
+ * by the slow-logging() call, when unpacked by the l3_dump.py script.
```
gapisback, line 41: Removed all references to the L3-logging APIs from comments as it was cofusing the grep / parsing in the helper pytest method, `build_exp_loc_msg_list()`.

use-cases/single-file-C-program/test-main.c
```diff
@@ -103,7 +113,7 @@ test_perf_fast_logging(int nMil)
     }
     uint32_t n = 0;
     for (n = 0; n < (nMil * L3_MILLION); n++) {
-        l3_log_fast("300-Mil Fast l3-log msgs", n, 0);
+        l3_log_fast("Perf-300-Mil Fast l3-log msgs", n, 0);
```
gapisback, line 116: We do not want to grep for these instances and on L93, above, so prefix the message with the term `Perf-`.

The grep code in `build_exp_loc_msg_list()` will filter these instances out.

l3_dump.py
```diff
@@ -227,8 +226,26 @@ def main():
...
+    Main method that drives the unpacking of the L3-dump file to print
+    diagnostic data. This modularized method exists outside of main() so that
+    it can be called independently via pytests.
+
+    Arguments:
+        args[0] - Str: Name of L3-log file
+        args[1] - Str: Name of binary that generated L3-log file.
+
+    Returns: Tuple:
+        # entries processed
```
gregthelaw, line 243: We should probably document the return tuple.

l3_dump.py
```diff
@@ -237,7 +254,15 @@ def main():
 
     strings = parse_rodata_string_offsets(program_bin)
 
-    with open(sys.argv[1], 'rb') as file:
+    # To enable unit-testing, via pytests, the parsing and return-data logic
```
gregthelaw, line 257: Are you saying that returning a bunch of lists, as opposed to a list of tuples, somehow makes it easier to unit test? It seems slightly odd to this. i.e. maintaining a list of tuples seems cleaner than a bunch of parallel lists.

tests/pytests/l3_dump_test.py
```diff
@@ -0,0 +1,545 @@
...
+import l3_dump
+
+L3DUMPSCRIPT    = 'l3_dump.py'
+L3Dump          = L3RootDir + '/' + L3DUMPSCRIPT
+
+L3_LOC_EXPLICITLY_UNSET = "0"
+L3_LOC_DEFAULT          = "1"
+L3_LOC_ELF_ENCODING     = "2"
+
+# Ternary: variable = something if condition else something_else
```
gregthelaw, line 39: is this comment supposed to be here?

tests/pytests/l3_dump_test.py
gregthelaw: Did you consider using pexpect here? Can be useful for this sort of testing.

use-cases/single-file-CC-program/test-main.cc
```diff
@@ -32,27 +34,33 @@ timespec_to_ns(struct timespec *ts)
...
 int
-main(void)
+main(const int argc, const char * argv[])
 {
-    // Logging perf benchmarking methods below share this log-file.
-    auto e = l3_init("/tmp/l3.cc-test.dat");
-    if (e) {
-        abort();
+    const char *logfile = NULL;
+    if (argc == 1) {
```
gregthelaw, line 40: Should we check the usage more carefully? If I'm reading this correctly, we can pass any command-line arguments at all and it will skip perf benchmarking. Which is OK I guess, but a little weird/confusing.

gregthelaw (approved):
Minor comments only - feel free to ignore.

l3_dump.py
gapisback: I, too, thought of returning a list of tuples when writing this first. But that got more difficult to work with for unit-testing.

As an example, see `test_unit_test_dump_log_entries()` in `pytests/l3_dump_test.py`.

One of the validations we do there is to create a hand-rolled expected-messages list and compare with the list output by this `do_main()` method.

```
 88     # Invoke the L3-dump script to unpack the slow-log-entries.
 89     (nentries, tid_list, loc_list, msg_list, arg1_list, arg2_list) \
 90         = l3_dump.do_main(['/tmp/l3.c-small-unit-test.dat', binary],
 91                           return_logentry_lists = True)
 92
 93     # -----------------------------------------------------------------------
 94     # Here is where it gets interesting. Hard-coded verification of expected
 95     # message and arguments, as unpacked by utility. This is one example of
 96     # verification using hard-coded expected-outputs lists.
 97     # -----------------------------------------------------------------------
 98     exp_msg_list = [  'Simple-log-msg-Args(1,2)'
 99                     , 'Simple-log-msg-Args(3,4)'
100                     , 'Potential memory overwrite (addr, size)'
101                     , 'Invalid buffer handle (addr)'
102                    ]
103     assert msg_list == exp_msg_list
```

It became easy to do L103; i.e., simply compare actual and expected list of messages.

If `do_main()` were to return a list of tuples, I would have to crack that open in this verifier to create the list of actual messages, i.e., `msg_list`.

I don't see the current method too too bad. Let me get this in, sit on it, and if I see benefits to returning a list of tuples, I can change all this easily enough.

l3_dump.py
gapisback: Done. Corrected the return type signature, as well.

tests/pytests/l3_dump_test.py
gapisback: Yes; I left behind as I continually find myself searching for this syntax, so left a small comment in-there.

use-cases/single-file-CC-program/test-main.cc
gapisback: Yes, perf benchmarking will only be run with no arguments.

I just wanted to suppress that when this program is invoked with `--unit-tests` argument. So changed the code in the 3 use-case sample programs to only do the `else()` part [i.e., non-perf unit-tests] if `--unit-tests` arg is provided.

gapisback:
@gregthelaw - I didn't know about the existence of `pexpect()` interface till you brought it up.

Took a quick look ... and seems like it's ideally suited for running commands in a child process to which you may want to communicate by sending inputs etc.

I suppose the new test code could be reconstructed to use `pexpect`, but I'd like to push that off to future. I don't want to spend the time now trying to figure out how to inject this paradigm, while other PRs are waiting behind this one to merge.

I'll track this suggestion as an open item to follow in our Discussions #23 work-items list.

PR #34, opened 2024-05-01
[Perf] Extend Client-server perf tests to run with L3_LOC_ENABLED

gapisback:
This PR packages two commits to integrate L3-logging invoked by the client-server performance test programs to now also invoke LOC-encoding as part of L3-Logging. The changes are mainly in `Makefile` with simple additions to `test.sh` and to `build.yml` to run these new changes thru CI.

1. [Perf] Makefile support to build client-server perf programs w/LOC encoding
2. [Perf] Makefile support to build client-server perf programs w/LOC-ELF

------

### 1. [Perf] Makefile support to build client-server perf programs w/LOC encoding

This commit implements Makefile rules to build the client-server programs such that L3-logging is performed with default LOC-encoding.

`./test.sh` : Add test-build-and-run-client-server-perf-test-l3_loc_eq_1 which is also invoked thru a call in CI build.yml

This test-case will run the client/server app and then will invoke the `l3_dump.py` to unpack the L3-log entries, with LOC-ID decoded.

As the LOC-generator, by default, spits out the LOC-decoder with the name of the dir/ it is searching, we need a way to specify the LOC-decoder binary to be used when dumping L3 log-entries.

`l3_dump.py` is now extended to support `--loc-binary` argument to specify the name of the LOC-decoder binary to use. (This feature is useful even by itself outside this very test-specific usage.)

To test this manually do:

```
$ make clean && CC=gcc LD=g++ BUILD_VERBOSE=1 L3_LOC_ENABLED=1 make client-server-perf-test

$ build/release/bin/use-cases/svmsg_file_server

$ build/release/bin/use-cases/svmsg_file_client 10

$ L3_LOC_ENABLED=1 ./l3_dump.py --log-file /tmp/l3.c-server-test.dat --binary build/release/bin/use-cases/svmsg_file_server --loc-binary build/release/bin/use-cases/client-server-msgs-perf_loc
```

----

### 2. [Perf] Makefile support to build client-server perf programs w/LOC-ELF encoding

This commit implements Makefile rules to build the client-server programs such that L3-logging is performed with enhanced LOC-ELF encoding.

`./test.sh` : Add `test-build-and-run-client-server-perf-test-l3_loc_eq_2`  which is also invoked thru a call in CI build.yml

This test-case will run the client/server app and then will invoke the `l3_dump.py` to unpack the L3-log entries. (Currently we still don't support decoding LOC-IDs encoded with LOC-ELF scheme.)

To test this manually do:

```
$ make clean && CC=gcc LD=g++ BUILD_VERBOSE=1 L3_LOC_ENABLED=2 make client-server-perf-test

$ build/release/bin/use-cases/svmsg_file_server

$ build/release/bin/use-cases/svmsg_file_client 10

$ L3_LOC_ENABLED=2 ./l3_dump.py --log-file /tmp/l3.c-server-test.dat --binary build/release/bin/use-cases/svmsg_file_server
```

### DIFF
--- a/l3_dump.py
+++ b/l3_dump.py
@@ -46,7 +46,7 @@ elif OS_UNAME_S == 'Darwin':
     READELF_DATA_SECTION = '__cstring'
 
 # #############################################################################
-def set_decode_loc_id(program_bin:str, loc_decoder_bin:str):
+def set_decode_loc_id(program_bin:str, loc_decoder_bin:str = None):
     """
     Check for required LOC-decoder binary, based on environment of run.
     If L3_LOC_ENABLED=1 is set in the env:
--- a/l3_dump.py
+++ b/l3_dump.py
@@ -46,18 +46,19 @@ elif OS_UNAME_S == 'Darwin':
     READELF_DATA_SECTION = '__cstring'
 
 # #############################################################################
-def set_decode_loc_id(program_bin:str):
+def set_decode_loc_id(program_bin:str, loc_decoder_bin:str):
     """
-    Check for required LOC-decoder binary, based on environment of run
-    If L3_LOC_ENABLED=1 is set in the env, search for a LOC-decoder binary
-    named "<program_binary>_loc".
+    Check for required LOC-decoder binary, based on environment of run.
+    If L3_LOC_ENABLED=1 is set in the env:
+     - Search for a LOC-decoder binary specified by the --loc-binary argument.
+     - Otherwiwse, search for a default LOC-binary named "<program_binary>_loc".
     """
     decode_loc_id = DECODE_LOC_ID
     loc_decoder = LOC_DECODER
     if LOC_ENABLED in os.environ:
         env_var_val = os.getenv(LOC_ENABLED)
         if env_var_val == "1":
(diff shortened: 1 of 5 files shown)


l3_dump.py
```diff
@@ -46,18 +46,19 @@
     READELF_DATA_SECTION = '__cstring'
 
 # #############################################################################
-def set_decode_loc_id(program_bin:str):
+def set_decode_loc_id(program_bin:str, loc_decoder_bin:str):
```
gregthelaw, line 49: Shouldn't `loc_decoder_bin`'s type be `str | None`?

gregthelaw (approved):
One minor comment, no need for further review/approval.

BTW, top-tip: you can change the PR branch you're merging into. So I changed it here from `master` to `msgsnd-rcv-perf` -- helps when queuing up multiple PRs when you have to deal with tardy reviewers like me :)

l3_dump.py
gapisback: Could be ... I changed it to:

```
 49 def set_decode_loc_id(program_bin:str, loc_decoder_bin:str = None):
```

... which is probably a bit more self-documenting.

PR #35, opened 2024-05-01
Formalize L3-log header. Add status flags. Remove dependence on env-variables.

gapisback:
Rework L3 log-file header to identify type of logging done.

This commit overhauls the L3_LOG{} header struct at the head of a L3-logfile. Several changes are introduced to make the handling of log-files a bit more robust, and independent of environment
variables.

l3.c:
     - Add enum l3_log_flags_t, to specify bit-flags to track type of logging in this log-file
     - Rearrange L3_LOG{} to include `l3_log_flags_t  flags` field.
     - Record the type of LOC-encoding used, which is determined as part of the build.

l3_dump.py:
     - Add l3_unpack_loghdr() to figure out the type of LOC-encoding present, if any, in the log-entries.
     - Rework identification of LOC-decode binary to use, without relying on environment variables.
        If the Log header indicates that `LOC_DEFAULT` is in effect (i.e., the application was built
       with `L3_LOC_ENABLED=1` env-var), then the dump utility will expect to find a corresponding
        LOC-decoder utility at a known location.
       If not, user has to supply one using the `--loc-binary` argument.
    - Remove dependency on os.environ() to figure out if LOC-decoding should be done.

l3_dump_test.py:
 - Cleaned-up test cases to no longer rely on OS env-variables to
   figure out the setting of L3_LOC_ENABLED when dumping a log-file.

 - With this change, existing verify_l3_dump_unpack() can nicely call
   into newly added l3_dump.l3_unpack_loghdr(), to figure the LOC-encoding
   scheme in-effect in log-file being dumped.

 - Fixed few minor testcase bugs.

-----

### NOTE about status flags added:

We record the platform (Linux, Mac/OSX) on which the logging was done. In future revs, this provides a place where we can cross-check that dumping on Mac/OSX a log-file that was generated on Linux, or vice-versa, can be checked-for and prevented.

Also, a magic field has been added, but it's not being used right now. In future, this will provide a place to add some magic number checking to ensure that `l3_dump.py` does not error out when presented  a file which is not a L3-log-file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -639,9 +639,6 @@ ifeq ($(L3_LOC_ENABLED), $(L3_LOC_DEFAULT))
 
 else ifeq ($(L3_LOC_ENABLED), $(L3_LOC_ELF_ENCODING))
 
-    # Core L3 files flag off of L3_LOC_ENABLED.
-    CFLAGS += -DL3_LOC_ENABLED
-
     # Core L3 files flag off of L3_LOC_ENABLED. L3_LOC_ELF_ENABLED is provided
     # in case in some sources we wish to conditionally compile some info-msgs
     # indicating that this LOC-ELF encoding scheme is in effect.
--- a/l3_dump.py
+++ b/l3_dump.py
@@ -22,11 +22,20 @@ L3_LOG_HEADER_SZ = 32  # bytes; offsetof(L3_LOG, slots)
 L3_ENTRY_SZ = 32       # bytes; sizeof(L3_ENTRY)
 
 # #############################################################################
-# #############################################################################
 PROGRAM_BIN = 'unknown'
 DECODE_LOC_ID = 0   # By default, we expect L3 logging was done w/LOC OFF.
 LOC_ENABLED = "L3_LOC_ENABLED"
 LOC_DECODER = "LOC_DECODER"
+
+# -----------------------------------------------------------------------------
+# The driving env-var, L3_LOC_ENABLED, needs to be set to one of these values.
+# LOC-encoding comes in two flavours. Default technique is based on the
+# Python-generator script. Enhanced technique is based on LOC-ELF
+# encoding.
+# -----------------------------------------------------------------------------
+L3_LOC_UNSET            = 0
+L3_LOC_DEFAULT          = 1
+L3_LOC_ELF_ENCODING     = 2
 
 # #############################################################################
 # Establish path / names to tools used here based on the OS-platform version
@@ -46,28 +55,11 @@ elif OS_UNAME_S == 'Darwin':
     READELF_DATA_SECTION = '__cstring'
 
 # #############################################################################
-def set_decode_loc_id(program_bin:str, loc_decoder_bin:str = None):
-    """
-    Check for required LOC-decoder binary, based on environment of run.
-    If L3_LOC_ENABLED=1 is set in the env:
-     - Search for a LOC-decoder binary specified by the --loc-binary argument.
-     - Otherwiwse, search for a default LOC-binary named "<program_binary>_loc".
-    """
-    decode_loc_id = DECODE_LOC_ID
-    loc_decoder = LOC_DECODER
-    if LOC_ENABLED in os.environ:
-        env_var_val = os.getenv(LOC_ENABLED)
-        if env_var_val == "1":
-            loc_decoder = program_bin + "_loc" if loc_decoder_bin is None else loc_decoder_bin
-            if os.path.exists(loc_decoder) is False:
-                print(f"Env-var {LOC_ENABLED}=1 is set, but required "
-                      + f"LOC-decoder binary {loc_decoder} is not found.")
-                sys.exit(1)
-
-            # LOC is in-use and LOC-decoder binary was found.
-            decode_loc_id = 1
-
-    return (decode_loc_id, loc_decoder)
+# Status flags defined in src/l3.c for L3_LOG()->flags field
+L3_LOG_FLAGS_LINUX                = int('0x0001', 16)
+L3_LOG_FLAGS_MACOSX               = int('0x0002', 16)
+L3_LOG_FLAGS_LOC_ENCODING         = int('0x0010', 16)
+L3_LOG_FLAGS_LOC_ELF_ENCODING     = int('0x0020', 16)
 
 # #############################################################################
 def which_binary(os_uname_s:str, bin_name:str):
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -639,9 +639,6 @@ ifeq ($(L3_LOC_ENABLED), $(L3_LOC_DEFAULT))
 
 else ifeq ($(L3_LOC_ENABLED), $(L3_LOC_ELF_ENCODING))
 
-    # Core L3 files flag off of L3_LOC_ENABLED.
-    CFLAGS += -DL3_LOC_ENABLED
```
gapisback, line 544: Deleted because it's repeated two lines below.

l3_dump.py
```diff
@@ -46,28 +55,11 @@
...
-
-            # LOC is in-use and LOC-decoder binary was found.
-            decode_loc_id = 1
-
-    return (decode_loc_id, loc_decoder)
+# Status flags defined in src/l3.c for L3_LOG()->flags field
+L3_LOG_FLAGS_LINUX                = int('0x0001', 16)
+L3_LOG_FLAGS_MACOSX               = int('0x0002', 16)
+L3_LOG_FLAGS_LOC_ENCODING         = int('0x0010', 16)
+L3_LOG_FLAGS_LOC_ELF_ENCODING     = int('0x0020', 16)
```
gapisback, line 62: Qs. Is there a more Pythonic way of definining these status flags?

This is the best I could come up with ... suggestions welcome.

Also, fyi -- this creates a dependency between these bits and their definitions in `l3.c`.

In future revs, I will look into having one set of definitions in some `.h` file and source them somehow in this `py` script.

src/l3.c
```diff
@@ -77,20 +84,48 @@ L3_STATIC_ASSERT(sizeof(loc_t) == sizeof(uint32_t),
...
-    uint64_t fbase_addr;
-    uint64_t pad0;
-    uint64_t pad1;
-    L3_ENTRY slots[L3_MAX_SLOTS];
+    uint64_t        idx;
+    uint64_t        fbase_addr;
+    uint32_t        pad0;
+    uint16_t        pad2;
+    l3_log_flags_t  flags;
+    uint64_t        pad1;
```
gapisback, line 113: Sizeof the struct has not changed. Just externalized some pad fields into `magic` (L110) and `flags` (L112).

tests/pytests/l3_dump_test.py
```diff
@@ -216,7 +216,7 @@ def test_cc_test_dump_log_entries():
     make_rv = exec_make(['make', 'all-cc-tests'],
                         { "BUILD_VERBOSE": "1",
                           "CC": "g++", "CXX": "g++", "LD": "g++",
-                          "L3_LOC_ENABLED": L3_LOC_DEFAULT} )
```
gapisback, line 219: This, and L191 above, were latent test-case bugs left-over from previous work.

These cases are supposed to test w/o LOC-encoding and supplying this env-var was incorrectly generate LOC-encoding as part of logging.

tests/pytests/l3_dump_test.py
```diff
@@ -264,7 +261,6 @@ def test_c_test_dump_log_entries_loc_eq_1():
                                       binary,
                                       usecase_prog_dir, 'test-main.c')
     assert unpack_rv is True
-    os.environ.pop('L3_LOC_ENABLED')
```
gapisback, line 267: This and L246 -- removes dependency on environment variables while dumping.

Dumping will figure out if LOC-encoding was active, and which scheme, by looking at the status flags recording the LOC-scheme.

l3_dump.py
gregthelaw: Looks Pythonic enough to me (although I'm not really an expert here.

However, I wonder whether a bit-mask of flags is the most appropriate here -- it cannot be Linux _and_ MacOS. So maybe there should be an OS code (0 = Linux, 1 = MacOS, 2 = Windows, etc), and separately an encoding type (0 = no LOC, 1 = regular LOC, 2 = ELF LOC). Rather than storing in a single bitmask.

(Just a thought, I don't feel super strongly about it.)

gregthelaw: Should we also store the number of log entry slots while we're here?

l3_dump.py
```diff
@@ -218,6 +210,69 @@ def parse_string_offsets(input_str:str) -> (dict,int):
...
+
+    typedef struct l3_log
+    {
+        uint64_t        idx;
+        uint64_t        fbase_addr;
+        uint32_t        pad0;
+        uint16_t        pad2;
+        l3_log_flags_t  flags;
+        uint64_t        pad1;
+        // L3_ENTRY        slots[L3_MAX_SLOTS];
```
gregthelaw, line 227: why is this commented out?

l3_dump.py
```diff
@@ -218,6 +210,69 @@ def parse_string_offsets(input_str:str) -> (dict,int):
...
+def l3_unpack_loghdr(file_hdl) -> (int, int):
+    """
+    Unpack the header struct of a L3-log file and identify key fields.
+    We are unpacking a struct laid out like the following:
+
+    typedef struct l3_log
+    {
+        uint64_t        idx;
+        uint64_t        fbase_addr;
+        uint32_t        pad0;
```
gregthelaw, line 223: rather than padding, should use `__attribute__((align(n)))` ?

l3_dump.py
```diff
@@ -218,6 +210,69 @@ def parse_string_offsets(input_str:str) -> (dict,int):
...
+    decode_loc_id, and the input program binary's name, figure out which
+    LOC-decoder binary to be used while unpacking LOC-ID values that may be
+    encoded in log-entries.
+
+    Currently, we only support decoding LOC-IDs encoded by default LOC scheme.
+     - Search for a LOC-decoder binary specified by the --loc-binary argument.
+     - Otherwiwse, search for a default LOC-binary named "<program_binary>_loc".
+    """
+    loc_decoder = None
+    if decode_loc_id == 1:
```
gregthelaw, line 266: Use `L3_LOC_DEFAULT` rather than `1`?

src/l3.c
gregthelaw: Use `__attribute__((align(n)))`?

gregthelaw (commented):
Looks good. I made some minor suggestions but none of them is a blocker, and no need to re-request review when you're done.

gregthelaw (approved):

l3_dump.py
gapisback: @gregthelaw -- I thought about this ... And You are right on both accounts.

I was just trying to save some precious log-header bytes so came up with bit-masks.

But, I agree, that's also potentially misleading as the platform bits are mutually-exclusive, just as are the LOC-encoding scheme bits.

Here is what I will do:

```
106 typedef struct l3_log
107 {
108     uint64_t        idx;
109     uint64_t        fbase_addr;
110     uint32_t        pad0;
111     uint16_t        ~pad2~ log_size;     // In # of log-entries; i.e. L3_MAX_SLOTS
112     ~l3_log_flags_t  flags;~
         BYTE              platform;
         BYTE.             loc_type;
113     uint64_t        pad1;
114     L3_ENTRY        slots[L3_MAX_SLOTS];
115 } L3_LOG;
```

----

I also thought of more compact representations, squeezing 4-bits for each `platform` and `loc_type` into 1-byte and then 1-byte to store `L3_MAX_SLOTS` as # of bits. This will save some bytes, but will restrict # of log-entries to be power-of-2. I don't immediately see a strong reason for this restriction.

Let's go with slightly liberal space-layout and can fine-tune as more development converges.

src/l3.c
gapisback: I'd like to stay with the scheme of externalizing pad-bytes added for compiler-padding as, then, it's apparent that some spare bytes are available for future development.

Also, I'm a bit leery of relying on compiler `__attribute__`s -- just in case those are not supported similarly on all compilers (e.g. clang, MS-Visual C++ etc.)

l3_dump.py
gapisback: I'd like to stay with the scheme of externalizing pad-bytes added for compiler-padding as, then, it's apparent that some spare bytes are available for future development.

Also, I'm a bit leery of relying on compiler __attribute__s -- just in case those are not supported similarly on all compilers (e.g. clang, MS-Visual C++ etc.)

l3_dump.py
gapisback: yes; done.

l3_dump.py
gapisback: Fixed; old cruft ...

PR #37, opened 2024-05-05
Gapisback/l3 unpack loghdr c printf sample programs

gapisback:


l3_dump.py
```diff
@@ -480,9 +480,21 @@ def do_main(args:list, return_logentry_lists:bool = False):
             # No location-ID will be recorded in log-files if L3_LOC_ENABLED is OFF.
             UNPACK_LOC = ''
             if loc == 0:
-                print(f"{tid=} '{msg_text}'")
+
```
gregthelaw, line 483: This change looks out of place - should it really be in this commit?

gregthelaw (changes requested):

l3_dump.py
gapisback: You are right. I discovered this 'error' while building some tests for these sample-programs.

I have applied this change as part of the previous commit which reworked the `L3_LOG{}` structure into types etc.

```
3ab13eb Aditya P. Gurajada 7 hours ago Mon, 06-May-2024, 05:24:22 PM (Authored: Mon, 06-May-2024, 12:01:18 PM)
 (L3) Rework L3_LOG{} struct to track platform, loc-encoding as types.
```

Above commit has the checks applied so they are now pulled-out of this PR / commit.

gregthelaw (approved):

PR #42, opened 2024-05-07
[Perf] Track thread-CPU-time for each request on server-side.

gapisback:
## Objective

The objective of this change is to demonstrate some realistic usages of L3-logging, which is currently done on the server-side of a simple client/server msg-send-recv application.

Rather than logging a plain 'Server msg' info, with some client details, we now also track the real-time / thread-CPU-time consumed on the server-side, for each client-op request. We only measure the block of code after msgrecv(), so don't track other scaffolding overheads.

The L3-logging call is now changed to also log this metric, which is a more natural use-case for this kind of logging API.

@gregthelaw - You can start reviewing this multi-part PR. I have reworked the implementation based on what we discussed yesterday and to prepare this change-set to be more inline with the `l3-perf-bm` proposal aired out in this discussion #43 . 

------

## Implementation

This PR has 3 commits in this order:

1. l3_dump.py: Add errors='ignore' clause to parse_rodata_string_offsets(): Fix applied to dump Python script to deal with some case where Unicode chars are generated in one of the test workloads.

2. [Perf] Track time-consumed for processing each request on server-side.

3. [Perf] Improve usability and metrics of client-server perf testbed: Tracks throughput metrics.

With the last 2 changes, following is the summary output w/o L3-logging (baseline) and with L3-logging:

```
agurajada-Linux-Vm:[3152] $ egrep -E 'Start Server|clients, num_ops' ~/tmp/perf.1Mil.out

Start Server, using clock ID=3 'CLOCK_THREAD_CPUTIME_ID': No logging.
For 5 clients, num_ops=5000000 (5 Million) ops, Avg. Thread-CPU time=542 ns/msg, throughput=1843753 (~1.84 Million) ops/sec

Start Server, using clock '/tmp/l3.c-server-test.dat': Initiate L3-logging to log-file '(no LOC)', using CLOCK_THREAD_CPUTIME_ID encoding scheme.
For 5 clients, num_ops=5000000 (5 Million) ops, Avg. Thread-CPU time=587 ns/msg, throughput=1703080 (~1.70 Million) ops/sec
```

-------

This commit implements on the server-side:

- Thread-specific CPU-time consumed for processing each client request received.
- Cumulative CPU-time and # requests is also tracked for each client.
- L3-log is changed to log the CPU-time consumed to implement each request.
- When client exits, the stats of # of requests received and avg. CPU-time/request is printed.

Measurement clock chosen:

 - Linux: Use CLOCK_REALTIME as it seems to have a resolution of 1 ns.
 - Mac: We use CLOCK_THREAD_CPUTIME_ID to measure the time taken to implement
   each request as its resolution is 1ns. All other clocks have a resolution
   of 1000 ns (as seen on my Mac).

With this enhancement, we should see these enhanced messages upon running the
L3-dump after running the exercise:

- Refactor `common timespec_to_ns()` into `svmsg_file.h` as it's now shared by client / server files.

We should see these enhanced messages (on a Mac/OSX):

- On server-side:
```
Server: Client ID=5701633 exited. num_ops=1000, Avg. thread-CPU-ns=351 ns/msg, # active clients=2
Server: Client ID=5570563 exited. num_ops=1000, Avg. thread-CPU-ns=351 ns/msg, # active clients=1
Client: ID=5570564 Performed 1000 message send/receive operations, ctr=1000, 35803 ns/msg (avg) Exiting.
Server: Client ID=5570564 exited. num_ops=1000, Avg. thread-CPU-ns=352 ns/msg, # active clients=0
Server: # active clients=0 (HWM=5). Exiting.
```

- From L3-dump utility:
```
tid=259 'Server msg: Increment: ClientID=5570563, Thread CPU-time=326 ns.' tid=259 'Server msg: Increment: ClientID=5570564, Thread CPU-time=361 ns.' tid=259 'Server msg: Increment: ClientID=5570564, Thread CPU-time=446 ns.' tid=259 'Server msg: Increment: ClientID=5570564, Thread CPU-time=453 ns.' Unpacked nentries=5000 log-entries.
```

---
With L3_LOC enabled L3-log'ged messages are like follows:

---
tid=259 client-server-msgs-perf/svmsg_file_server.c:258  'Server msg: Increment: ClientID=5636100, Thread CPU-time=333 ns.'
tid=259 client-server-msgs-perf/svmsg_file_server.c:258  'Server msg: Increment: ClientID=5636098, Thread CPU-time=331 ns.'
tid=259 client-server-msgs-perf/svmsg_file_server.c:258  'Server msg: Increment: ClientID=5767169, Thread CPU-time=334 ns.'
tid=259 client-server-msgs-perf/svmsg_file_server.c:258  'Server msg: Increment: ClientID=5636100, Thread CPU-time=329 ns.'
tid=259 client-server-msgs-perf/svmsg_file_server.c:258  'Server msg: Increment: ClientID=5767169, Thread CPU-time=446 ns.'
tid=259 client-server-msgs-perf/svmsg_file_server.c:258  'Server msg: Increment: ClientID=5636100, Thread CPU-time=340 ns.'
tid=259 client-server-msgs-perf/svmsg_file_server.c:258  'Server msg: Increment: ClientID=5767169, Thread CPU-time=425 ns.'
Unpacked nentries=5000 log-entries.

--
Note the `Thread CPU-time=340 ns.` fragment, which is new with this change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ help::
 	@echo ' make run-cc-tests'
 	@echo ' '
 	@echo 'To build client-server performance test programs and run performance test'
-	@echo ' make clean && CXX=g++ LD=g++ make client-server-perf-test'
+	@echo ' make clean && CC=gcc LD=g++ make client-server-perf-test'
 	@echo ' '
 	@echo 'Environment variables: '
 	@echo ' BUILD_MODE={release,debug}'
--- a/Makefile
+++ b/Makefile
@@ -48,8 +48,11 @@ help::
 	@echo ' make clean && CC=g++ CXX=g++ LD=g++ make all-cc-tests'
 	@echo ' make run-cc-tests'
 	@echo ' '
-	@echo 'To build client-server performance test programs and run performance test'
-	@echo ' make clean && CC=gcc LD=g++ make client-server-perf-test'
+	@echo 'To build client-server performance test programs and run performance test(s)'
+	@echo ' make clean && CC=gcc LD=g++ L3_ENABLED=0     make client-server-perf-test  # Baseline'
+	@echo ' make clean && CC=gcc LD=g++                  make client-server-perf-test  # L3-logging'
+	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=1 make client-server-perf-test  # L3+LOC logging'
+	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=2 make client-server-perf-test  # L3+LOC-ELF logging'
 	@echo ' '
 	@echo 'Environment variables: '
 	@echo ' BUILD_MODE={release,debug}'
@@ -553,6 +556,8 @@ CLIENT_SERVER_PERF_TESTS_DIR   := $(USE_CASES)/client-server-msgs-perf
 # we will build standalone client and server binaries.
 CLIENT_SERVER_PERF_TESTS_SRCS   := $(wildcard $(CLIENT_SERVER_PERF_TESTS_DIR)/*.c)
 
+CLIENT_SERVER_PERF_TESTS_SRCS   += $(L3_UTILS_SRCS)
+
 # Build pair of sources for client-/server-main sources.
 CLIENT_SERVER_CLIENT_MAIN_SRC   := $(filter %_client.c, $(CLIENT_SERVER_PERF_TESTS_SRCS))
 CLIENT_SERVER_SERVER_MAIN_SRC   := $(filter %_server.c, $(CLIENT_SERVER_PERF_TESTS_SRCS))
@@ -613,6 +618,7 @@ $(CLIENT_SERVER_PROGRAM_GENSRC): | $(BINDIR)/$(USE_CASES)/.
 	@$(LOCGENPY) --gen-includes-dir  $(L3_INCDIR) --gen-source-dir $(dir $@) --src-root-dir $(dir $@) --loc-decoder-dir $(BINDIR)/$(USE_CASES) --verbose
 	@echo
 
+client-server-perf-test: INCLUDE += -I ./$(L3_UTILSDIR)
 client-server-perf-test: $(CLIENT_SERVER_PROGRAM_GENSRC) $(CLIENT_SERVER_PERF_TEST_BINS)
 
 # ##############################################################################
--- a/use-cases/client-server-msgs-perf/svmsg_file_server.c
+++ b/use-cases/client-server-msgs-perf/svmsg_file_server.c
@@ -45,6 +45,7 @@
    serially.
 */
 #include <inttypes.h>
+#include <time.h>
 #include <assert.h>
 #include "svmsg_file.h"
 #include "l3.h"
@@ -56,6 +57,9 @@ typedef struct client_info {
     int             clientId;
     int             client_idx; // Into ActiveClients[] array, below
     int64_t         client_ctr; // Current value of client's counter
+    uint64_t        cumu_thread_cpu_time_ns;
+                                // Cumulative thread-CPU time for operation.
+    uint64_t        num_ops;    // # of msg-op requests received
     req_resp_type_t last_mtype; // Last requests->mtype
 } Client_info;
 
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -49,7 +49,7 @@ help::
 	@echo ' make run-cc-tests'
 	@echo ' '
 	@echo 'To build client-server performance test programs and run performance test'
-	@echo ' make clean && CXX=g++ LD=g++ make client-server-perf-test'
+	@echo ' make clean && CC=gcc LD=g++ make client-server-perf-test'
```
gapisback, line 52: Minor correction. We just need `gcc` for these programs for now; no need for `g++`.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -56,6 +57,9 @@ typedef struct client_info {
     int             clientId;
     int             client_idx; // Into ActiveClients[] array, below
     int64_t         client_ctr; // Current value of client's counter
+    uint64_t        cumu_time_ns;
+                                // Cumulative time consumed for operation.
+    uint64_t        num_ops;    // # of msg-op requests received
```
gapisback, line 62: Added these two metrics.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -90,13 +94,50 @@ grimReaper(int sig)
     errno = savedErrno;
 }
 
+// Find the resolution of the clock used in the program's timing measurement.
+void
+svr_clock_getres(void)
```
gapisback, line 99: Added to find out resolution of diff clocks.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -174,32 +266,82 @@ main(int argc, char *argv[])
...
+            // Record time-consumed. (NOTE: See clarification below.)
+            l3_log_simple("Server msg: Increment: ClientID=%d, "
+                          "Thread-CPU-time=%" PRIu64 " ns.",
+                          resp.clientId, elapsed_ns);
 #endif // L3_ENABLED
+
+            // Reuse variable to reacquire current TS for accumulation.
+            // Cumulative time-consumed gives a read into impact of any
+            // logging call that might be done in diff micro-benchmarks.
+            if (clock_gettime(clock_id, &ts1)) {    // Cumulative timing ends
```
gapisback, line 307: This metric of cumulative time for the operation, which includes logging call on L299, will give us the overhead incurred by L3-logging.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -190,16 +252,49 @@ main(int argc, char *argv[])
...
 
           case REQ_MT_EXIT:
+
+            clientp = &ActiveClients[req.client_idx];
             // One client has informed that it has exited
             NumActiveClients--;
-            printf("Server: Client ID=%d exited. # active clients=%d\n",
-                   req.clientId, NumActiveClients);
+
+#if __APPLE__
```
gregthelaw: minor: you could do this once and use above.

gregthelaw (approved):
Looks good. One minor suggestion, although if we're not supporting Mac for now it doesn't really matter.

I 100% agree with the approach of not supporting Mac. Let's get some traction with this, if there is some decent traction we can add Mac support then.

FWIW I'm quite disappointed that L3 adds so much overhead. It's essentially an atomic add and a few stores -- I would have expected it not to be measurable. Especially as it affects transactions per second so much -- I would have expected the RPC/context-switch time to completely dominate here. I guess we need to measure against `fprintf` and spdlog to know for sure, but something seems a bit odd.

use-cases/client-server-msgs-perf/svmsg_file_server.c
gapisback: @gregthelaw - This code fragment is now obsolete. 

I had tried to get this working for Apple, hence, this messy code.

As I have backed-out support for Mac/OSX for now, all this has been obsoleted by a later commit.

gapisback:
@gregthelaw - thanks for looking at this work.

Re: "_FWIW I'm quite disappointed that L3 adds so much overhead_" ... in [this comment](https://github.com/undoio/l3/pull/42#pullrequestreview-2051429233), here's my take.

I look at this change-set as a stable working implementation on which we can do more u-benchmarking experiments.

Do note that I have only run L3-logging in "slow" mode, i.e., it's a function call to a C-function, not the assembly code. I'll see if I can somehow stitch a perf-version where one can select slow / fast logging api.

I am running this on a Linux-VM running on my Mac, so there probably are other overheads.

Also, the perf-experimentations for `fprintf()`, spdlog will happen soon, but for that I'd like this core infra piece checked-in.

I think once we get this work stable, it will be good to port this to real h/w or a VM running on some cloud which may have more stable performance behaviour. (Let's take this offline.)

I'll give this a final look and will prepare for check-in to `/main`.

PR #44, opened 2024-05-10
[Util] Add size_to_str() and value_to_str() unit-conversion interfaces

gapisback:
This commit adds following size-/value to string with unit-specifier interfaces.

- Convert 'size' value (e.g., memory, space) to a string: 
```
char * size_to_str(char *outbuf, size_t outbuflen, size_t size);
char * size_to_fmtstr(char *outbuf, size_t outbuflen, const char *fmtstr, size_t size);
```

- Convert number-value value (e.g. # of requests) to a string:
```
char * value_to_str(char *outbuf, size_t outbuflen, size_t size);
char * value_to_fmtstr(char *outbuf, size_t outbuflen, const char *fmtstr, size_t size);
```
The size_to_str() implementation is extracted, and made stand-alone, out of the implementation done in the [SplinterDB repo](https://github.com/vmware/splinterdb). 

Further extended to add the `value_to_str()` interfaces.

- Add unit-test tests/unit/size_str-test.c
- Add Makefile rules to build and run unit-test.
- Fixes to get new code and tests working on Mac/OSX.

### Sample outputs:

  - Size  = 2048 (2 KiB)            Value = 2048 (~2.04 K)
  - Size  = 1000000 (~976.56 KiB)   Value = 1000000 (1 Million)
  - Size  = 1073741952 (~1.00 GiB)  Value = 1073741952 (~1.07 Billion)
  - Size  = 1099511627776 (1 TiB)   Value = 1099511627776 (~1.09 Trillion)

### Note to the reviewer:

@gregthelaw -- I am assigning a review to you, just so you see this change-set coming-in.

These are utility methods outside core L3 sources and I need these for improving the info-msgs from the performance test exerciser. Hence, I plan to check these in. The comments should be self-explanatory. And this code has been implemented and tested elsewhere, so I am comfortable about its correctness.

If you have any suggestions to improve this helper method, happy to take that up in a later change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,6 +113,8 @@ INCDIR      := include
 L3_SRCDIR   := $(SRCDIR)
 L3_INCDIR   := $(INCDIR)
 USE_CASES   := use-cases
+L3_UTILS    := utils
+L3_UTILSDIR := $(USE_CASES)/$(L3_UTILS)
 
 # Name of L3 dump utility that unpacks L3 log-entries
 L3_DUMP                 := l3_dump.py
@@ -220,6 +222,9 @@ endif
 
 L3_SRCS     := $(L3_SRC) $(L3_ASSEMBLY)
 L3_OBJS     := $(L3_SRCS:%.c=$(OBJDIR)/%.o)
+
+L3_UTILS_SRCS   := $(L3_UTILSDIR)/size_str.c
+L3_UTILS_OBJS   := $(L3_UTILS_SRCS:%.c=$(OBJDIR)/%.o)
 
 # Symbol for all unit-test sources, from which we will build standalone
 # unit-test binaries.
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -113,6 +113,8 @@ INCDIR      := include
 L3_SRCDIR   := $(SRCDIR)
 L3_INCDIR   := $(INCDIR)
 USE_CASES   := use-cases
+L3_UTILS    := utils
+L3_UTILSDIR := $(USE_CASES)/$(L3_UTILS)
```
gapisback, line 117: Am adding the new helper functions under `use-cases/utils/` as I next plan to use the `size_str()` interfaces in the `use-cases/client-server-msgs-perf` programs.

These helper conversion methods are not core-L3 functionality, hence, rather than littering a top-level `utils/` sub-dir, I chose to stash them under the `use-cases/utils/` dir.

If we do find more wide-spread for these helper conversion functions, we can look at relocating the sources at some more visible upper-level dirs.

Makefile
```diff
@@ -428,7 +434,7 @@ endif
 # bin/unit/<x>: obj/unit/<x>.o
 define unit_test_self_dependency =
 $(1): $(patsubst $(BINDIR)/$(UNIT_DIR)/%,$(OBJDIR)/$(UNITTESTS_DIR)/%.o, $(1)) \
-        $(L3_OBJS)
+        $(L3_OBJS) $(L3_UTILS_OBJS)
```
gapisback, line 437: Also run the new unit-test binary.

PR #47, opened 2024-05-13
[L3] Rework logging APIs to invoke printf() for fmt-spec validation

gapisback:
This commit lays down the groundwork to re-engage u-benchmarking
programs to compare the overheads of L3-logging v/s, say, logging
via `printf()` or `write()` system call. 

In order to be able to do that, we first need to validate that the print format-specifiers in the 'msg' supplied to l3_log*() are consistent with the arguments supplied.

Currently, several of the use-case and sample programs are coded loosely w/o adhering to this requirement.

This commit implements the following changes:

 - Slighly rearranges calling sequence of L3-logging APIs.
 - Defines DEBUG-version of the caller-macros, which under an `else if (0)` [ dead-code block ] will evaluate to a call to `printf()`.
 - This will trip-up compilation errors in debug-mode builds if any use-case of L3-logging apis have a message with print format-specifiers that are inconsistent with the args supplied.

Requiring that the args-supplied are consistent with the print-format specifiers in the `msg` also makes the job of L3-dump utility lot more reliable and simpler. (It will also help in future
when we do support var-args to the L3-logging apis.)

Bulk of the changes are in `include/l3.h` with minor updates to test programs to conform to the new calling requirements. We still retain required 2-args to the L3-logging call. (Var-args support
will be considered in a future change.)

-----

**NOTE: There are no functional enhancements / changes. Only internal interface refactoring.**

### DIFF
--- a/include/l3.h
+++ b/include/l3.h
@@ -55,14 +55,57 @@ int l3_init(const char *path);
 
 /**
  * \brief Caller-macro to invoke L3 simple logging.
+ *
+ * Macro to synthesize LOC-encoding value, if needed, under conditional
+ * compilation flags. DEBUG-version of macros are provided to cross-check
+ * that the supplied format string in `msg` are consistent with the
+ * arguments' type.
  */
-#ifdef L3_LOC_ENABLED
-#define l3_log_simple(msg, arg1, arg2)                          \
-        l3__log_simple(__LOC__, (msg), (arg1), (arg2))
-#else   // L3_LOC_ENABLED
-#define l3_log_simple(msg, arg1, arg2)                          \
-        l3__log_simple(L3_ARG_UNUSED, (msg), (arg1), (arg2))
-#endif  //
+#if defined(DEBUG)
+
+  #ifdef L3_LOC_ENABLED
+
+    #define l3_log_simple(msg, arg1, arg2)                              \
+            if (1) {                                                    \
+                l3__log_simple((msg),                                   \
+                               (uint64_t) (arg1), (uint64_t) (arg2),    \
+                                 __LOC__);                              \
+            } else if (0) {                                             \
+                printf((msg), (arg1), (arg2));                          \
+            } else
+
+   #else   // L3_LOC_ENABLED
+
+    #define l3_log_simple(msg, arg1, arg2)                              \
+            if (1) {                                                    \
+                l3__log_simple((msg),                                   \
+                               (uint64_t) (arg1), (uint64_t) (arg2),    \
+                               L3_ARG_UNUSED);                          \
+            } else if (0) {                                             \
+                printf((msg), (arg1), (arg2));                          \
+            } else
+
+  #endif  // L3_LOC_ENABLED
+
+#else   // defined(DEBUG)
+
+  #ifdef L3_LOC_ENABLED
+
+    #define l3_log_simple(msg, arg1, arg2)                          \
+            l3__log_simple((msg),                                   \
+                           (uint64_t) (arg1), (uint64_t) (arg2),    \
+                           __LOC__)
+
+  #else   // L3_LOC_ENABLED
+
+    #define l3_log_simple(msg, arg1, arg2)                          \
+            l3__log_simple((msg),                                   \
+                           (uint64_t) (arg1), (uint64_t) (arg2),    \
+                           L3_ARG_UNUSED)
+
+  #endif  // L3_LOC_ENABLED
+
+#endif  // defined(DEBUG)
 
 /**
  * \brief Log a literal string and two arguments.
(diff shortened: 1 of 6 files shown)


include/l3.h
```diff
@@ -55,14 +55,57 @@ int l3_init(const char *path);
...
+ * arguments' type.
  */
-#ifdef L3_LOC_ENABLED
-#define l3_log_simple(msg, arg1, arg2)                          \
-        l3__log_simple(__LOC__, (msg), (arg1), (arg2))
-#else   // L3_LOC_ENABLED
-#define l3_log_simple(msg, arg1, arg2)                          \
-        l3__log_simple(L3_ARG_UNUSED, (msg), (arg1), (arg2))
-#endif  //
+#if defined(DEBUG)
```
gapisback, line 64: This is a pattern you will see in this change-set.

Redefine caller-macros under `DEBUG` build mode, so that we eventually compile `printf()`, which will trip-up if calling code is wrong.

include/l3.h
```diff
@@ -55,14 +55,57 @@ int l3_init(const char *path);
...
+
+  #ifdef L3_LOC_ENABLED
+
+    #define l3_log_simple(msg, arg1, arg2)                              \
+            if (1) {                                                    \
+                l3__log_simple((msg),                                   \
+                               (uint64_t) (arg1), (uint64_t) (arg2),    \
+                                 __LOC__);                              \
+            } else if (0) {                                             \
+                printf((msg), (arg1), (arg2));                          \
```
gapisback, line 74: This is where we leverage the compiler, and L73 is dead-code.

include/l3.h
```diff
@@ -55,14 +55,57 @@ int l3_init(const char *path);
...
+            } else
```
gapisback, line 75: To ensure that a call to the macro gets rightly terminated as : `if () ... else;`

include/l3.h
```diff
@@ -81,11 +124,11 @@ extern "C" {
 #endif
 
 #ifdef L3_LOC_ENABLED
-void l3__log_simple(const loc_t loc, const char *msg,
-                    const uint64_t arg1, const uint64_t arg2);
+void l3__log_simple(const char *msg, const uint64_t arg1, const uint64_t arg2,
+                    const loc_t loc);
```
gapisback, line 128: Arg rearrangement was needed so that we can invoke `printf()` with just :

`const char *msg, const uint64_t arg1, const uint64_t arg`

as printf() does not know aything about `loc_t loc` argument.

include/l3.h
gregthelaw: You can use `__attribute__((format(printf, x, y)))` which will invoke the compiler warnings. Might be a bit simpler and potentially give better compiler errors? Supported by both gcc and clang.

include/l3.h
gregthelaw: again, the `format` attribute allows for this.

include/l3.h
gapisback: Thanks for this tip, @gregthelaw -- I learnt something new today!

I tried this out ... but found out thru experimentation, and after reading [the docs carefully](https://gcc.gnu.org/onlinedocs/gcc-3.2/gcc/Function-Attributes.html), that this syntax is valid only if the argument lists are defined as varargs, using `...`.

See this doc note:


format (archetype, string-index, first-to-check)
    The format attribute specifies that a function takes printf, scanf, strftime or strfmon style arguments which should be type-checked against a format string. For example, the declaration:

              extern int
              my_printf (void *my_object, const char *my_format, ...)
                    __attribute__ ((format (printf, 2, 3)));
              

    causes the compiler to check the arguments in calls to my_printf for consistency with the printf style format string argument my_format. 
    
    ---

NOTE that the example shows: `const char *my_format, ...)` ... which means var-args are being used.

----

When I tried to implement this attribute, I get this compilation error:

```
gcc  -x c -DL3_ENABLED -D_GNU_SOURCE -ggdb3 -Wall -Wfatal-errors -Werror -I ./include -I ./tests/unit/ -c tests/unit/l3_dump.py-test.c -o build/release/obj/tests/unit/l3_dump.py-test.o
In file included from tests/unit/l3_dump.py-test.c:27:
./include/l3.h:131:21: error: ‘format’ attribute argument 3 value ‘2’ does not refer to a variable argument list
  131 |                     const uint32_t loc) __attribute__ ((format(__printf__, 1, 2)));
      |                     ^~~~~
compilation terminated due to -Wfatal-errors.
```

----

I resisted converting the log-msg interface to use var-args right now, in this change-set.

That extension is something I have in mind, in a future change, once we get the rest of core perf-benchmarking working.

gapisback:
@gregthelaw : I think we should make a separate pass to re-define `l3_log()` interfaces to take-in `(const char *msg, ...)` varargs notation.

At that time, your suggestion to use `__attribute__ format(printf, ...)` can be incorporated.

If you are ok with the rest of the changes and can pass the review, I can check this in to `/main.`

include/l3.h
gregthelaw: Ahh, makes sense!

gregthelaw (approved):

PR #48, opened 2024-05-15
[L3] Rename l3__log_simple() -> l3_log_mmap()

gapisback:
This commit renames lower-level `l3__log_simple()` -> `l3_log_mmap()`, to indicate that the default L3-logging API is to an `mmap()`'ed log-file. There are no other functional changes introduced with this commit.


gapisback:
FYI, @gregthelaw - The renaming change we spoke about. I'm checking this in, but you may wish to give this a quick check and sign-off.

PR #49, opened 2024-05-15
[L3] Introduce l3_log_init() as a shell to call l3_init().

gapisback:
This commit lays down some boilerplate code to add support for different logging schemes under L3. Introduce enum l3_log_t to specify different logging interfaces, mainly intended for use by micro-benchmarking utility programs. l3_log_init() becomes a shell around l3_init(), which remains the default interface to `init` a L3-logging system, on `mmap()`-ed files.

There are no other functional changes introduced with this commit.

### DIFF
--- a/include/l3.h
+++ b/include/l3.h
@@ -47,6 +47,18 @@
 extern "C" {
 #endif
 
+/**
+ * Different logging interfaces, mainly intended for use by micro-benchmarking
+ * utility programs.
+ */
+typedef enum {
+      L3_LOG_UNDEF      = 0
+    , L3_LOG_MMAP
+    , L3_LOG_FPRINTF
+    , L3_LOG_DEFAULT    = L3_LOG_MMAP
+} l3_log_t;
+
+int l3_log_init(const l3_log_t logtype, const char *path);
 int l3_init(const char *path);
 
 #ifdef __cplusplus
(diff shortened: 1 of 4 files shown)


include/l3.h
```diff
@@ -47,6 +47,18 @@
...
 #endif
 
+/**
+ * Different logging interfaces, mainly intended for use by micro-benchmarking
+ * utility programs.
+ */
+typedef enum {
+      L3_LOG_UNDEF      = 0
+    , L3_LOG_MMAP
+    , L3_LOG_FPRINTF
```
gapisback, line 57: Adding this to lay down the ground-work to extend existing client-server performance tests to run with diff logging schemes.

We may, well, choose to take this out later, but for now, this change seems useful to enable further dev of perf u-benchmarking programs.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -192,7 +192,7 @@ main(int argc, char *argv[])
     // Initialize L3-Logging
 #if L3_ENABLED
     const char *logfile = "/tmp/l3.c-server-test.dat";
-    int e = l3_init(logfile);
+    int e = l3_log_init(L3_LOG_MMAP, logfile);
```
gapisback, line 195: Currently, I only need this performance benchmarking program to tap-into `l3_log_init()`, so it can be extended to other logging-schemes.

I am leaving the other call-sites which use `l3_init()` untouched at their default mode of mmap-logging. 
That change would needlessly introduce more code-diffs, for this kind of interface change that may not survive in future (tbd).

gapisback:
FYI, @gregthelaw - Another renaming change we spoke about. I'm checking this in, but you may wish to give this a quick check and sign-off.

PR #53, opened 2024-05-16
[Perf] Extend client-server perf test to invoke L3-fast logging

gapisback:
This commit adds a new perf-test execution mode for the client-server performance u-benchmarking toolkit.

`L3_FASTLOG_ENABLED=1 make client-server-perf-test` will build the programs so that server will invoke l3_log_fast(), when needed.

test.sh: Refactoring.
  - Extend test-build-and-run-client-server-perf-test to also execute a case when L3-fast-logging is invoked.
  - Cleanup: Use local variables instead of hard-coded boolean 0/1s to specify if L3/LOC is enabled  / disabled.

svmsg_file_server.c: Invoke l3_log_fast() under `#if L3_FASTLOG_ENABLED`
  which is supplied by the Makefile build-rules.

### Usage:

Build client/server programs to perform L3-fast logging:

```
$ make clean && CC=gcc LD=g++ L3_FASTLOG_ENABLED=1 make client-server-perf-test
```
### Perf u-benchmark results (Linux-VM on Mac): 5 clients, 1Mil messages:

**Units**: Throughput measured on server-side, in terms of Million ops/sec
```
   Workload         Expt-1     Expt-2
   Baseline         ~1.96       ~2.00
   L3-logging       ~1.94       ~1.82
   L3-fast logging  ~1.86       ~1.92
```
The 2nd experiment shows the expected results. There seems to be some fluctuations in these runs, probably owning to the VM.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,10 +49,11 @@ help::
 	@echo ' make run-cc-tests'
 	@echo ' '
 	@echo 'To build client-server performance test programs and run performance test(s)'
-	@echo ' make clean && CC=gcc LD=g++ L3_ENABLED=0     make client-server-perf-test  # Baseline'
-	@echo ' make clean && CC=gcc LD=g++                  make client-server-perf-test  # L3-logging'
-	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=1 make client-server-perf-test  # L3+LOC logging'
-	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=2 make client-server-perf-test  # L3+LOC-ELF logging'
+	@echo ' make clean && CC=gcc LD=g++ L3_ENABLED=0         make client-server-perf-test  # Baseline'
+	@echo ' make clean && CC=gcc LD=g++                      make client-server-perf-test  # L3-logging'
+	@echo ' make clean && CC=gcc LD=g++ L3_FASTLOG_ENABLED=1 make client-server-perf-test  # L3 Fast logging'
+	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=1     make client-server-perf-test  # L3+LOC logging'
+	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=2     make client-server-perf-test  # L3+LOC-ELF logging'
 	@echo ' '
 	@echo 'Environment variables: '
 	@echo ' BUILD_MODE={release,debug}'
@@ -568,6 +569,13 @@ CLIENT_SERVER_NON_MAIN_SRCS     := $(filter-out $(CLIENT_SERVER_PERF_TESTS_DIR)/
 # If L3-logging is enabled in test-program, need to include core library file
 ifeq ($(L3_ENABLED), $(L3_DEFAULT))
     CLIENT_SERVER_NON_MAIN_SRCS += $(L3_SRC)
+
+# Under L3-logging, invoke the L3-fast logging API, which needs the .S file
+ifeq ($(L3_FASTLOG_ENABLED), 1)
+    CLIENT_SERVER_NON_MAIN_SRCS += $(L3_ASSEMBLY)
+    CFLAGS += -DL3_FASTLOG_ENABLED
+endif
+
 endif
 
 # Name of LOC-generated source file for the client-server perf-test program.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -49,10 +49,11 @@ help::
 	@echo ' make run-cc-tests'
 	@echo ' '
 	@echo 'To build client-server performance test programs and run performance test(s)'
-	@echo ' make clean && CC=gcc LD=g++ L3_ENABLED=0     make client-server-perf-test  # Baseline'
-	@echo ' make clean && CC=gcc LD=g++                  make client-server-perf-test  # L3-logging'
-	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=1 make client-server-perf-test  # L3+LOC logging'
-	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=2 make client-server-perf-test  # L3+LOC-ELF logging'
+	@echo ' make clean && CC=gcc LD=g++ L3_ENABLED=0         make client-server-perf-test  # Baseline'
+	@echo ' make clean && CC=gcc LD=g++                      make client-server-perf-test  # L3-logging'
+	@echo ' make clean && CC=gcc LD=g++ L3_FASTLOG_ENABLED=1 make client-server-perf-test  # L3 Fast logging'
```
gapisback, line 54: This is the new build mode.

The rest of the edits are to indent the help output aligned in columnar-outputs.

test.sh
```diff
@@ -393,15 +393,41 @@ function test-build-and-run-client-server-perf-test()
...
+    local l3_log_disabled=0
+    local l3_log_enabled=1
+    local l3_LOC_disabled=0
+
     echo " "
     echo "${Me}: Client-server performance testing with L3-logging OFF:"
     echo " "
-    build-and-run-client-server-perf-test "${num_msgs_per_client}" 0
+    build-and-run-client-server-perf-test "${num_msgs_per_client}"      \
+                                          "${l3_log_disabled}"
```
gapisback, line 404: Local variable replaces use of hard-coded boolean `0`.

test.sh
```diff
@@ -393,15 +393,41 @@ function test-build-and-run-client-server-perf-test()
...
+            --binary "./build/${Build_mode}/bin/use-cases/svmsg_file_server"    \
+        | tail -${nentries}
+
+    echo " "
+    echo "${Me}: Client-server performance testing with L3-fast-logging ON:"
+    echo " "
+    build-and-run-client-server-perf-test "${num_msgs_per_client}"  \
+                                          "${l3_log_enabled}"       \
+                                          "${l3_LOC_disabled}"      \
+                                          "fast"
```
gapisback, line 430: We want to run the exercise with L3-logging enabled, without tracking extra LOC-ID, -but- we do want fast logging ON.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -295,10 +302,18 @@ main(int argc, char *argv[])
             elapsed_ns = (nsec1 - nsec0);           // Elapsed-ns for this op
 
 #if L3_ENABLED
+
             // Record time-consumed. (NOTE: See clarification below.)
+#  if L3_FASTLOG_ENABLED
+            l3_log_fast("Server msg: Increment: ClientID=%d, "
+                        "Thread-CPU-time=%" PRIu64 " ns. (L3-fast-log)",
```
gapisback, line 309: @gregthelaw - This is a mystery.

All the right env-vars and `-D flags` are being passed thru the build system.

I changed the message to say `(L3-fast-log)`, intentionally.

And I manually verified in CI and my dev-runs that this term **is coming** out, from L3-dump utility. 

So **L308 is being invoked**, but how come the overall throughput metric in some experiments is  slower for L3-fast v/s L3-slow logging?

That's very strange. I would have expected that with this change the overall throughput metric for this new case will be as good as, or better than, the plain L3-logging.

All this perf measurements on my Linux-VM is becoming unreliable. We really need to run these tests on your Linux laptop directly.

gregthelaw: Yes, I too would have expected l3-fast to be faster! Maybe the compiler generates code that is more VM-friendly. Who knows? Will have to try it on my native Linux laptop to see.

gregthelaw (approved):

PR #57, opened 2024-05-22
[L3] Rename l3_log_simple() -> l3_log(). No other diffs.

gapisback:
Replace references in pytest code, also.

### DIFF
--- a/include/l3.h
+++ b/include/l3.h
@@ -66,7 +66,7 @@ int l3_init(const char *path);
 #endif
 
 /**
- * \brief Caller-macro to invoke L3 simple logging.
+ * \brief Caller-macro to invoke L3 logging.
  *
  * Macro to synthesize LOC-encoding value, if needed, under conditional
  * compilation flags. DEBUG-version of macros are provided to cross-check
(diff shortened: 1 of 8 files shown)


include/l3.h
```diff
@@ -77,7 +77,7 @@ int l3_init(const char *path);
 
   #ifdef L3_LOC_ENABLED
 
-    #define l3_log_simple(msg, arg1, arg2)                              \
+    #define l3_log(msg, arg1, arg2)                                     \
```
gapisback, line 80: The core change is simply this renaming. `l3_log()` will, henceforth, be the external logging interface.

Previous commits have already introduced `l3_log_mmap()` as the lower-level interface, so this upper-level renaming falls into place nicely.

Rest of the changes in this change-set are simply refactoring name change.

gapisback:
@gregthelaw - Assigned this review, as an fyi. It's a minor code-cleanup; something we talked about earlier.

Let me know if you find this `l3_log()` external interface totally unacceptable.

gregthelaw:
Makes total sense. At some point I think we should make `l3_log()` just always use the asm implementation, but no rush on that.

PR #58, opened 2024-05-23
[Perf] Revise 'time-consumed' tracking on server-side.

gapisback:
This commit simplifies the time-tracking fields for each client. Previously, we tried to track two different times:

 1. CPU-clock-time for implementing the actual op
 2. Real-time-clock-time for the whole process including the overheads of logging.

The idea was to measure (1) [ w/o logging ] and to track it as one of the arguments to the subsequent l3_log() call.

Another attempt made was to convert the cumulative [2] time to a throughput metric (num-ops / sec). But trying to mix these two time-tracking schemes, and jostle between `clock_gettime()` calls got confusing and needlessly unreliable.

This commit simplifies everything as follows.

Implement the code as:

```
  a) clock_gettime() ◁----------------------------+
  b) Do the n-operations in a forever-loop        |--> time measured for
  c) Log, if needed [ Absent in baseline runs ]   | configured clock
  d) clock_gettime() ◁----------------------------+
```

This commit reworks the time-tracking fields for each side of the transmission.  We just use the elapsed time between (d) and (a) to compute the overall throughput.

**Client**: Get CLOCK_REALTIME for the entire n-messages and generate client-side throughput. This will measure
  the impact of L3-logging on real-time throughput.

  When client-exits, report back its throughput to the server.

**Server**: Use CLOCK_REALTIME as default clock.  Start timer before forever-loop waiting for messsges from
  the clients. End the timer when all clients exit.

  Assuming that clients are continually pumping messages, this is a reasonable way to measure overall elapsed-time.

  Server will report avg throughput across all clients.

The server program has already been extended to support cmdline flags to select the type of clock to use.

So, with this support, one can run the server as `svmsg_file_server --clock-process-cputime-id` to run through the
u-benchmarking using the CLOCK_PROCESS_CPUTIME_ID clock, to get a read-into the timing metrics using that clock. Default clock is `CLOCK_REALTIME`.

test.sh: Driver script is enhanced to run various perf-bm tests, varying
  the run-parameters. See `./test.sh --help` for detailed usage info.
  Fix shellcheck errors.

- Added svr_clock_calibrate() & svr_clock_overhead() to run thru clock calibration on test machine.

- Improved / expanded msgs reporting metrics / summary.

With this rework, all perf-test-methods will report the client-and server-side throughput using real-time elapsed, aggregating across c-clients and n-messages.

Tests can be run using, say, --clock-process-cputime-id flag, to generate the overhead due to logging in terms of CPU-time spent when L3 is enabled. Throughput reported on server-side is also scaled based on CPU-time-ns elapsed.

-----

### NOTE TO THE REVIEWER

@gregthelaw -- I have updated the instructions in the [L3‐Performance-Benchmarking]() Wiki page to reflect the interface changes introduced by this change-set. 

You should be able to download this dev-branch and follow those instructions to re-run these u-benchmarking tests on your Linux workstation.

-----

#### Results from latest iteration of fix:

5-concurrent clients, sending 1 Million messages to the server:

- Using real-time clock on both client and server-side:

```
agurajada-Linux-Vm:[195] $ egrep -E 'Server throughput' ~/tmp/run-all-perf.client-svr-throughput.1Mil.5ct.out| cut -f2,6,7 -d','
 No logging, Server throughput=98569 (~98.56 K) ops/sec, Client throughput=23569 (~23.56 K) ops/sec
 L3-logging (no LOC), Server throughput=100628 (~100.62 K) ops/sec, Client throughput=22690 (~22.69 K) ops/sec
 L3-fast logging (no LOC), Server throughput=103407 (~103.40 K) ops/sec, Client throughput=24236 (~24.23 K) ops/sec
 L3-logging default LOC, Server throughput=103843 (~103.84 K) ops/sec, Client throughput=25804 (~25.80 K) ops/sec
 L3-logging LOC-ELF, Server throughput=103263 (~103.26 K) ops/sec, Client throughput=23565 (~23.56 K) ops/sec
```

- Using real-time clock on client-side and CPU-time clock on server-side:

```
agurajada-Linux-Vm:[196] $ egrep -E 'Server throughput' ~/tmp/run-all-perf.client-svr-throughput.CPU-time.1Mil.5ct.out | cut -f2,6,7 -d','
 No logging, Server throughput=122093 (~122.09 K) ops/sec, Client throughput=23415 (~23.41 K) ops/sec
 L3-logging (no LOC), Server throughput=127401 (~127.40 K) ops/sec, Client throughput=24266 (~24.26 K) ops/sec
 L3-fast logging (no LOC), Server throughput=121472 (~121.47 K) ops/sec, Client throughput=23918 (~23.91 K) ops/sec
 L3-logging default LOC, Server throughput=122455 (~122.45 K) ops/sec, Client throughput=23655 (~23.65 K) ops/sec
 L3-logging LOC-ELF, Server throughput=129966 (~129.96 K) ops/sec, Client throughput=25562 (~25.56 K) ops/sec
```

The diff in throughput comparing baseline with L3-logging enabled is inline with expected ranges of 2-4% drop.

In some runs, the throughput with L3-logging enabled goes up, but this may be an aberration of timing computations and due to the use of the VM.

------

#### Results from previous iteration of fix (can be discarded eventually):

Here is a sample output on my Linux-VM:

- Throughput metrics using the CLOCK_REALTIME clock:

```
agurajada-Linux-Vm:[122] $ grep "msg, throughput=" ~/tmp/run-all-perf.defaults.1Mil.5ct.out | cut -f2,6 -d','
 No logging, throughput=33991082 (~33.99 Million) ops/sec
 L3-logging (no LOC), throughput=21455900 (~21.45 Million) ops/sec
 L3-fast logging (no LOC), throughput=18753585 (~18.75 Million) ops/sec
 L3-logging default LOC, throughput=20511098 (~20.51 Million) ops/sec
 L3-logging LOC-ELF, throughput=18114519 (~18.11 Million) ops/sec
```

Drop from ~33.99 to ~21.45 million ops/sec w/L3-logging enable: Drop of ~37%.

- Throughput metrics using the CLOCK_PROCESS_CPUTIME_ID clock:

```
agurajada-Linux-Vm:[119] $ grep "msg, throughput=" ~/tmp/run-all-perf.process-CPU-time.1Mil.5ct.out | cut -f2,6 -d','
 No logging, throughput=2983442 (~2.98 Million) ops/sec
 L3-logging (no LOC), throughput=2840394 (~2.84 Million) ops/sec
 L3-fast logging (no LOC), throughput=2900842 (~2.90 Million) ops/sec
 L3-logging default LOC, throughput=2995347 (~2.99 Million) ops/sec
 L3-logging LOC-ELF, throughput=2682040 (~2.68 Million) ops/sec
```

Drop from ~2.98 to ~2.84 million ops/sec w/L3-logging enable: Drop of ~4.6%, which is "in-line" with the drop expected w/L3 enabled.

**Help**: The puzzling thing for me is why the throughput metric drops so much when computed using real-time. I suppose that's to be expected, but would like to get your eyes to review that there are no code / logic / computation errors.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -39,8 +39,18 @@ function cleanup()
 
 trap cleanup ERR
 
+# ##################################################################
+# Global variables to support different workloads / configurations to
+# execute client-server performance u-benchmarking tests.
+# ##################################################################
+# Number of messages sent by each client to the server
+NumMsgsPerClient=1000
+
 # Number of clients started for client-server RPC-perf testing.
 NumClients=5
+
+# Server-clock cmdline arg. Default is CLOCK_REALTIME clock
+SvrClockArg=
 
 # ##################################################################
 # Array of test function names. If you add a new test_<function>,
@@ -103,6 +113,25 @@ TestList=(
 function usage()
 {
    echo "Usage: $Me [--help | --list] [ --from-test <test-function-name> ] [test_all]"
+
+   echo " "
+   echo "Run client-server performance tests as follows:"
+   echo " "
+   echo " ${Me} run-all-client-server-perf-tests [ server-clock-ID [ num-msgs [ num-clients ] ] ]"
+   echo " "
+   echo " ${Me} test-build-and-run-client-server-perf-test [ num-msgs [ server-clock-ID ] ]"
+   echo " "
+   echo " ${Me} test-build-and-run-client-server-perf-test-l3_loc_eq_1 [ num-msgs [ server-clock-ID ] ]"
+   echo " "
+   echo " ${Me} test-build-and-run-client-server-perf-test-l3_loc_eq_2 [ num-msgs [ server-clock-ID ] ]"
+   echo " "
+   echo "  Client-Server performance u-benchmarking defaults:"
+   echo "    Number of clients                      = ${NumClients}"
+   echo "    Number of messages sent by each client = ${NumMsgsPerClient}"
+   echo "    Default clock-ID used (client)         = CLOCK_REALTIME"
+   echo "    Default clock-ID used (server)         = CLOCK_REALTIME"
+   echo "    Alternate server clock-selector: See \`svmsg_file_server --help\`"
+   echo "       One of: --clock-default | --clock-monotonic | --clock-process-cputime-id | --clock-realtime | --clock-thread-cputime-id"
 }
 
 # --------------------------------------------------------------------------
@@ -352,24 +381,30 @@ function build-and-run-sample-c-appln-with-LOC-encoding()
 # Driver method to run through all variations of client-server perf tests
 #
 # Parameters:
-#   $1  - (Opt) # of messages to exchange from client -> server (default: 1000)
-#   $2  - (Opt) # of clients to start-up (default: 5)
+#   $1  - (Opt) Arg to select clock-ID to use
+#   $2  - (Opt) # of messages to exchange from client -> server (default: 1000)
+#   $3  - (Opt) # of clients to start-up (default: 5)
 #
 # Usage:
-#  $ ./test.sh run-all-client-server-perf-tests $((100 * 1000)) > /tmp/perf-test.100K-rows.5ct.out 2>&1
+#  $ ./test.sh run-all-client-server-perf-tests --clock-default $((100 * 1000)) > /tmp/perf-test.100K-rows.5ct.out 2>&1
 #
 #  $ egrep -B1 -E 'Start Server|, num_ops=|Unpacked nentries|throughput=' /tmp/perf-test.100K-rows.5ct.out | egrep -v -E 'Client-throughput|enabled on server-side.'
 
 # #############################################################################
 function run-all-client-server-perf-tests()
 {
-    local num_msgs_per_client=1000
+    # Reset global if arg-1 provided. Minions will grab it from there.
     if [ $# -ge 1 ]; then
-        num_msgs_per_client=$1
+        SvrClockArg=$1
     fi
 
-    if [ $# -eq 2 ]; then
-        NumClients=$2
+    local num_msgs_per_client=${NumMsgsPerClient}
+    if [ $# -ge 2 ]; then
+        num_msgs_per_client=$2
+    fi
+
+    if [ $# -ge 3 ]; then
+        NumClients=$3
     fi
 
     test-build-and-run-client-server-perf-test "${num_msgs_per_client}"
@@ -389,9 +424,13 @@ function test-build-and-run-client-server-perf-test()
         return
     fi
 
-    local num_msgs_per_client=1000
-    if [ $# -eq 1 ]; then
+    local num_msgs_per_client=${NumMsgsPerClient}
+    if [ $# -ge 1 ]; then
         num_msgs_per_client=$1
+    fi
+
+    if [ $# -ge 2 ]; then
+        SvrClockArg=$2
     fi
 
     local l3_log_disabled=0
@@ -399,13 +438,13 @@ function test-build-and-run-client-server-perf-test()
     local l3_LOC_disabled=0
 
     echo " "
-    echo "${Me}: Client-server performance testing with L3-logging OFF:"
+    echo "${Me}: Client-server performance testing with L3-logging OFF ${SvrClockArg}:"
     echo " "
     build-and-run-client-server-perf-test "${num_msgs_per_client}"      \
                                           "${l3_log_disabled}"
 
     echo " "
-    echo "${Me}: Client-server performance testing with L3-logging ON:"
+    echo "${Me}: Client-server performance testing with L3-logging ON ${SvrClockArg}:"
     echo " "
     build-and-run-client-server-perf-test "${num_msgs_per_client}" \
                                           "${l3_log_enabled}"
@@ -423,7 +462,7 @@ function test-build-and-run-client-server-perf-test()
         | tail -${nentries}
 
     echo " "
-    echo "${Me}: Client-server performance testing with L3-fast-logging ON:"
+    echo "${Me}: Client-server performance testing with L3-fast-logging ON ${SvrClockArg}:"
     echo " "
     build-and-run-client-server-perf-test "${num_msgs_per_client}"  \
                                           "${l3_log_enabled}"       \
@@ -467,16 +506,20 @@ function test-build-and-run-client-server-perf-test-l3_loc_eq_1()
         return
     fi
 
-    local num_msgs_per_client=1000
-    if [ $# -eq 1 ]; then
+    local num_msgs_per_client=${NumMsgsPerClient}
+    if [ $# -ge 1 ]; then
         num_msgs_per_client=$1
+    fi
+
+    if [ $# -ge 2 ]; then
+        SvrClockArg=$2
     fi
 
     local l3_log_enabled=1
     local l3_LOC_enabled=1
 
     echo " "
-    echo "${Me}: Client-server performance testing with L3-logging and L3-LOC ON:"
+    echo "${Me}: Client-server performance testing with L3-logging and L3-LOC ON ${SvrClockArg}:"
     echo " "
     build-and-run-client-server-perf-test "${num_msgs_per_client}"  \
                                           "${l3_log_enabled}"       \
@@ -506,22 +549,26 @@ function test-build-and-run-client-server-perf-test-l3_loc_eq_1()
 # #############################################################################
 function test-build-and-run-client-server-perf-test-l3_loc_eq_2()
 {
-    local num_msgs_per_client=1000
-    if [ $# -eq 1 ]; then
-        num_msgs_per_client=$1
-    fi
     set +x
-
     if [ "${UNAME_S}" = "Darwin" ]; then
         echo "${Me}: Client-server performance tests not supported currently on Mac/OSX."
         return
+    fi
+
+    local num_msgs_per_client=${NumMsgsPerClient}
+    if [ $# -ge 1 ]; then
+        num_msgs_per_client=$1
+    fi
+
+    if [ $# -ge 2 ]; then
+        SvrClockArg=$2
     fi
 
     local l3_log_enabled=1
     local l3_LOC_enabled=2
 
     echo " "
-    echo "${Me}: Client-server performance testing with L3-logging and LOC-ELF ON:"
+    echo "${Me}: Client-server performance testing with L3-logging and LOC-ELF ON ${SvrClockArg}:"
     echo " "
     build-and-run-client-server-perf-test "${num_msgs_per_client}"  \
                                           "${l3_log_enabled}"       \
@@ -604,7 +651,10 @@ function build-and-run-client-server-perf-test()
         esac
     fi
 
-    ${server_bin} &
+    # User can execute perf-bm tests with different server-clocks,
+    # using the --clock-<...> argument.
+    # shellcheck disable=SC2086
+    ${server_bin} ${SvrClockArg} &
 
     set +x
     sleep 5
--- a/use-cases/client-server-msgs-perf/svmsg_file_server.c
+++ b/use-cases/client-server-msgs-perf/svmsg_file_server.c
@@ -166,28 +166,22 @@ main(int argc, char *argv[])
     int     serverId;
     struct sigaction sa;
 
-    // On MacOSX, only the CLOCK_THREAD_CPUTIME_ID clock has a resolution of 1ns.
-    // On Linux, all clocks seem to have 1 ns resolution, so, default is to also
-    // use the same CLOCK_THREAD_CPUTIME_ID.
-    int clock_id;
-
 #if __APPLE__
     printf("%s is currently not supported on Mac/OSX\n", argv[0]);
     exit(EXIT_SUCCESS);
 #endif
 
-    // Default, so throughput (nops/sec) is consistent with normal expectation.
-    clock_id = CLOCK_REALTIME;
+    // On MacOSX, only the CLOCK_THREAD_CPUTIME_ID clock has a resolution of 1ns.
+    // On Linux, all clocks seem to have 1 ns resolution, so, default is to
+    // use the CLOCK_REALTIME, so throughput (nops/sec) is metris's calculation
+    // is consistent with normal expectation.
+    int clock_id = CLOCK_REALTIME;
 
     // Arg-parsing is only supported on Linux.
     int rv = parse_arguments(argc, argv, &clock_id);
     if (rv) {
         errExit("Argument error.");
     }
-
-#if !defined(L3_ENABLED)
-    svr_clock_calibrate();
-#endif  // L3_ENABLED
 
     /* Create server message queue */
 
@@ -401,6 +395,12 @@ end_forever_loop:
 
     printSummaryStats(run_descr, ActiveClients, NumActiveClientsHWM, clock_id,
                       elapsed_ns);
+
+    // For visibility into how clocks are performing on user's machine,
+    // run clock-calibration after all workload / metrics collection is done.
+#if !defined(L3_ENABLED)
+    svr_clock_calibrate();
+#endif  // L3_ENABLED
 
     exit(EXIT_SUCCESS);
 }
--- a/use-cases/client-server-msgs-perf/svmsg_file_server.c
+++ b/use-cases/client-server-msgs-perf/svmsg_file_server.c
@@ -19,7 +19,8 @@
    https://man7.org/tlpi/code/online/all_files_by_chapter.html
 
 Usage: ./svmsg_file_server --help
-       ./svmsg_file_server [ --clock-monotonic
+       ./svmsg_file_server [  --clock-default
+                            | --clock-monotonic
                             | --clock-realtime
                             | --clock-process-cputime-id
                             | --clock-thread-cputime-id ]
@@ -74,14 +75,21 @@ Usage: ./svmsg_file_server --help
 
 /**
  * Global data structures to track client-specific information.
+ *
+ * NOTE: This program can be configured to use different clocks to measure
+ * time, using the --clock<arg> argument. The cumu_time_ns tracks the
+ * cumulative time elapsed to implement the msg, including the time for
+ * any logging that might have been done in a build/experiment.
  */
 typedef struct client_info {
     int             clientId;
-    int             client_idx;     // Into ActiveClients[] array, below
-    int64_t         client_ctr;     // Current value of client's counter
-    uint64_t        cumu_time_ns;   // Cumulative time for operation.
-    uint64_t        num_ops;        // # of msg-op requests received
-    req_resp_type_t last_mtype;     // Last requests->mtype
+    int             client_idx;         // Into ActiveClients[] array, below
+    int64_t         client_ctr;         // Current value of client's counter
+    uint64_t        cumu_time_ns;       // Cumulative elapsed-time for operation,
+                                        // including logging overheads.
+    uint64_t        num_ops;            // # of msg-op requests received
+    uint64_t        throughput;         // Client-side avg throughput nops/sec
+    req_resp_type_t last_mtype;         // Last request's->mtype
 } Client_info;
 
 Client_info ActiveClients[MAX_CLIENTS];
@@ -107,7 +115,8 @@ static unsigned int   NumActiveClients    = 0;
  * Simple argument parsing structure.
  */
 struct option Long_options[] = {
-      { "help"                      , no_argument         , NULL, 'h'}
+      { "clock-default"             , no_argument         , NULL, 'd'}
+    , { "help"                      , no_argument         , NULL, 'h'}
     , { "clock-monotonic"           , no_argument         , NULL, 'm'}
     , { "clock-process-cputime-id"  , no_argument         , NULL, 'p'}
     , { "clock-realtime"            , no_argument         , NULL, 'r'}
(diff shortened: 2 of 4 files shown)


test.sh
```diff
@@ -39,9 +39,19 @@ function cleanup()
...
+# execute client-server performance u-benchmarking tests.
+# ##################################################################
+# Number of messages sent by each client to the server
+NumMsgsPerClient=1000
+
 # Number of clients started for client-server RPC-perf testing.
 NumClients=5
 
+# Server-clock cmdline arg. Default is CLOCK_REALTIME clock
+SvrClockArg=
```
gapisback, line 53: Some of the perf test-methods allow you to run `test.sh` overriding these defaults.

test.sh
```diff
@@ -352,24 +380,30 @@ function build-and-run-sample-c-appln-with-LOC-encoding()
...
 #  $ egrep -B1 -E 'Start Server|, num_ops=|Unpacked nentries|throughput=' /tmp/perf-test.100K-rows.5ct.out | egrep -v -E 'Client-throughput|enabled on server-side.'
 
 # #############################################################################
 function run-all-client-server-perf-tests()
 {
-    local num_msgs_per_client=1000
+    # Reset global if arg-1 provided. Minions will grab it from there.
     if [ $# -ge 1 ]; then
-        num_msgs_per_client=$1
+        SvrClockArg=$1
```
gapisback, line 397: Minor rearrangement of args to this method to simplify common-case execution of `test.sh` with another clock.

E.g. `$ ./test.sh run-all-client-server-perf-tests --clock-process-cputime-id`

to run a small workload consisting of 5-clients and 1000 msgts, with above run-parameters.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -107,24 +114,29 @@ static unsigned int   NumActiveClients    = 0;
...
+    , { "help"                      , no_argument         , NULL, 'h'}
     , { "clock-monotonic"           , no_argument         , NULL, 'm'}
     , { "clock-process-cputime-id"  , no_argument         , NULL, 'p'}
     , { "clock-realtime"            , no_argument         , NULL, 'r'}
     , { "clock-thread-cputime-id"   , no_argument         , NULL, 't'}
     , { NULL, 0, NULL, 0}           // End of options
 };
 
-const char * Options_str = ":hrmptd";
+const char * Options_str = ":dhmprt";
```
gapisback, line 126: Rearranged to match the order in which options are listed in above lookup array.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -219,19 +236,24 @@ main(int argc, char *argv[])
     printf("Start Server, using clock '%s'"
            ": Initiate L3-%slogging to log-file '%s'"
            ", using %s encoding scheme.\n",
-           logfile, l3_log_mode, loc_scheme, clock_name(clock_id));
+           clock_name(clock_id), l3_log_mode, logfile, loc_scheme);
+
+    snprintf(run_descr, sizeof(run_descr), "L3-%slogging %s",
+             l3_log_mode, loc_scheme);
```
gapisback, line 242: Generate run-description string, so it can be printed in summary output line.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -293,40 +313,33 @@ main(int argc, char *argv[])
...
-            // Cumulative time-consumed gives a read into impact of any
-            // logging call that might be done in diff micro-benchmarks.
-
-            if (clock_gettime(clock_id, &ts1)) {    // Cumulative timing ends
+            if (clock_gettime(clock_id, &ts1)) {        // **** Timing ends
                 errExit("clock_gettime-ts1");
             }
-            nsec1 = timespec_to_ns(&ts1);
+            uint64_t nsec0 = timespec_to_ns(&ts0);
+            uint64_t nsec1 = timespec_to_ns(&ts1);
```
gapisback, line 339: Do all timing-math -after- timing ends.

gapisback:
@gregthelaw -- the best way to verify if the perf #s / metrics are coming out ok is to install this dev-branch on your Linux workstation and to re-run the scripts.

The output files have itemized as well as aggregated metrics, so there is a TON of data for you to go thru to see if you can spot any errors, or to sign-off that the measurements' technique seems kosher.

Thanks!

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -293,41 +305,55 @@ main(int argc, char *argv[])
...
+                uint64_t elapsed_realtime_ns = (nsec1 - nsec0);
+
+                // Cumulative real-timeelapsed-ns
+                clientp->cumu_real_time_ns += elapsed_realtime_ns;
             }
-            nsec1 = timespec_to_ns(&ts1);
-            elapsed_ns = (nsec1 - nsec0);
-            clientp->cumu_time_ns += elapsed_ns;    // Cumulative elapsed-ns
 
+            clientp->realtime_clock ^= 1;
```
gregthelaw: what is this doing?

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -293,41 +305,55 @@ main(int argc, char *argv[])
...
-                errExit("clock_gettime-ts1");
+            if (clientp->realtime_clock) {
+                if (clock_gettime(tmp_clock_id, &ts1)) {    // Cumulative real-timing ends
+                    errExit("clock_gettime-ts1");
+                }
+                nsec1 = timespec_to_ns(&ts1);
+                uint64_t elapsed_realtime_ns = (nsec1 - nsec0);
+
+                // Cumulative real-timeelapsed-ns
+                clientp->cumu_real_time_ns += elapsed_realtime_ns;
```
gregthelaw: I don't follow the rationale here - can you explain a bit more? I would have thought the most (only?) reliable way is to do e.g. 1,000,000 transactions, and gettime once at the beginning and again at the end. If you do it like this, with a pair of gettime operations for each transaction, small errors/inaccuracies are going to accumulate, right? Also you'll also be including the time taken to call `clock_gettime()`. Sometimes that might result in a syscall which will then dominate.

use-cases/client-server-msgs-perf/svmsg_file_server.c
gapisback: ignore it, Greg -- That's an outdated diff, from a prior version of attempts to use diff clocks for timing. This line is now deleted.

use-cases/client-server-msgs-perf/svmsg_file_server.c
gapisback: @gregthelaw -- I'm sorry, but seems like I mislead you.

You are looking at an older commit which has been outdated by [this commit](https://github.com/undoio/l3/pull/58/commits/8790cde73135906de196bec76445fa991a8cec5a).

Forget this realtime tracking business.

I've described the whole approach in the top-level description of the PR. Let's take the discussion to the outer level of UI comments.

gapisback:
@gregthelaw -- recording your feedback sent to me privately by email:

I commented on the PR. FWIW I see _even worse_ results on my system!

```
No logging, throughput=21545873 (~21.54 Million) ops/sec
L3-logging (no LOC), throughput=4307118 (~4.30 Million) ops/sec
L3-fast logging (no LOC), throughput=4462253 (~4.46 Million) ops/sec
L3-logging default LOC, throughput=4776621 (~4.77 Million) ops/sec
L3-logging LOC-ELF, throughput=5110483 (~5.11 Million) ops/sec
No logging, throughput=27060377 (~27.06 Million) ops/sec
L3-logging (no LOC), throughput=4939081 (~4.93 Million) ops/sec
L3-fast logging (no LOC), throughput=5035875 (~5.03 Million) ops/sec
L3-logging default LOC, throughput=5247206 (~5.24 Million) ops/sec
L3-logging LOC-ELF, throughput=4390879 (~4.39 Million) ops/sec
```
i.e. logging drops the throughput by 80%!!

I think the problem is likely to be that a pair of clock_gettime()'s close together will give a wide error margin, and then you add them all up and the errors accumulate.

gapisback:
@gregthelaw : What this u-benchmark is trying to do is:

1. Client sends a simple message to server
2. Server "implements" the message (increment operator). Baseline involves no logging.
3. Add a L3-log call on server side for each message. Measure how much overhead a call to l3-logging would add.

That's why I "book-end" the `case` block on the server-side for the INCR operator with `clock_gettime()` calls, to get the before / after / elapsed time for [2].

About the overhead in the pair of nearby `clock_gettime()` calls, I argued to myself that that should be the same for baseline as it is for L3-logging case. But, seemingly it's not -- and that's also what is puzzling: Why?

---

About this overhead due to system call, I added the `svr_clock_calibrate()` method, which estimates an avg. overhead for a `clock_gettime()` system call. 

I did try an experiment where I reduced this estimated overhead when computing the overall elapsed_time_ns, like here:

```
335             if (clock_gettime(clock_id, &ts1)) {        // **** Timing ends
336                 errExit("clock_gettime-ts1");
337             }
338             uint64_t nsec0 = timespec_to_ns(&ts0);
339             uint64_t nsec1 = timespec_to_ns(&ts1);
340             elapsed_ns = (nsec1 - nsec0);
341
342             clientp->cumu_time_ns += elapsed_ns;   // Track Cumulative time
```

What you'd do is on L342, deduct (2 * estimated-overhead-of-system-call) to get something "closer" to actual elapsed-time-ns (for whichever clock is in use) for the actual of body of code implementing the message.

But, that also did not affect the computed throughput metric for baseline v/s L3-logging case.

---

We talked about delivering a "throughput" metric, so that is better represented by number-of-ops / real-time-clock-sec, and not using elapsed_time_ns as cpu-time-ns. That would lead to a misleading interpretation, I think, of throughput metric.

gregthelaw:
> About the overhead in the pair of nearby clock_gettime() calls, I argued to myself that that should be the same for baseline as it is for L3-logging case. But, seemingly it's not -- and that's also what is puzzling: Why?

Say that `clock_gettime()` takes an average of 10ns, and `l3_log()` also takes 10ns. A pair of `clock_gettime()`'s like:
```
clock_gettime(clock_id, &ts0);
clock_gettime(clock_id, &ts1);
```
will show, on average, 10ns between `ts0` and `ts1`. Then we add an `l3_log()`, like:
```
clock_gettime(clock_id, &ts0);
l3_log("hello world");
clock_gettime(clock_id, &ts1);
```
and it will show, on average, 20ns. If we just accumulate those times, we'll conclude that the addition of the `l3_log()` has halved the throughput. But that's totally misleading, because that time is trivial compared to the round-trip of the IPC. In practice, the round trip send/recv probably takes, say, 1ms, so by adding a single 10ns `l3_log()` we've actually reduced the throughput by 1%

Now, you've done the cailbration to remove the time taken by `clock_gettime()`, so effectively you're measuring the "throughput" of:
```
            resp.mtype = REQ_MT_INCR;               // Construct response message
            resp.clientId = req.clientId;           // For this client
            resp.client_idx = req.client_idx;

            clientp->last_mtype = REQ_MT_INCR;      // Prepare to increment
            resp.counter = ++clientp->client_ctr;   // Do Increment
            clientp->num_ops++;                     // Track # of operations
```
to 
```
            resp.mtype = REQ_MT_INCR;               // Construct response message
            resp.clientId = req.clientId;           // For this client
            resp.client_idx = req.client_idx;

            clientp->last_mtype = REQ_MT_INCR;      // Prepare to increment
            resp.counter = ++clientp->client_ctr;   // Do Increment
            clientp->num_ops++;                     // Track # of operations

            l3_log("Server msg: Increment: ClientID=%d, "
                   "Elapsed time=%" PRIu64 " ns.",
                   resp.clientId, elapsed_ns);
```

which is five assignments and two increments, with or without a log message. So yeah, if you're measuring that you'll find the addition of the log message significantly reduces the throughput, because it's, say, 10ns for the log message and 10ns for the assignments and increments. i.e. a handful of instructions for each.

But the client/server round trip will take something like a milisecond.

Hence I think the only meaningful measurement is the number of transactions per second. As you say, you can't count the transactions from the client because there are many clients, but there is only one server so you can count them there, right?

gapisback:
@gregthelaw - I have revised the entire time-tracking and throughput computation methodology.

The description is also updated, so you should re-read that to understand what the new does.

New results are posted in the main description pane.

To use this, on your Linux, you can do:

- Run with default clock-selection on both sides:

```
./test.sh run-all-client-server-perf-tests --clock-default $((1000 * 1000)) > ~/tmp/run-all-perf.client-svr-throughput.1Mil.5ct.out 2>&1
```

- Run with default clock-selection for client and CPU-time-clock-ID for server:
 
```
./test.sh run-all-client-server-perf-tests --clock-process-cputime-id  $((1000 * 1000)) > ~/tmp/run-all-perf.client-svr-throughput.CPU-time.1Mil.5ct.out 2>&1
```

- Post-process output as:

```
egrep -E 'Server throughput' ~/tmp/run-all-perf.client-svr-throughput.1Mil.5ct.out

egrep -E 'Server throughput' ~/tmp/run-all-perf.client-svr-throughput.1Mil.5ct.out| cut -f2,6,7 -d','
```

gregthelaw:
This looks much better!
```
gel@onyx:~/src/l3$ egrep -E 'Server throughput' ~/tmp/run-all-perf.client-svr-throughput.1Mil.5ct.out| cut -f2,6,7 -d',' No logging, Server throughput=512189 (~512.18 K) ops/sec, Client throughput=186824 (~186.82 K) ops/sec
 L3-logging (no LOC), Server throughput=480790 (~480.79 K) ops/sec, Client throughput=186319 (~186.31 K) ops/sec
 L3-fast logging (no LOC), Server throughput=480820 (~480.82 K) ops/sec, Client throughput=186962 (~186.96 K) ops/sec
 L3-logging default LOC, Server throughput=476344 (~476.34 K) ops/sec, Client throughput=184531 (~184.53 K) ops/sec
 L3-logging LOC-ELF, Server throughput=472868 (~472.86 K) ops/sec, Client throughput=181075 (~181.07 K) ops/sec
gel@onyx:~/src/l3$ 
```
-ve slowdown for the L3-fast logging :).

gregthelaw:
Oh, I misread, -ve slowdown on the client throughput, but significant slowdown on server throughput -- 8%. Looking at the code I can't see how the server throughput can be so different from the client throughput -- server just counts the sum of the ops from each client, right?

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -162,15 +176,18 @@ main(int argc, char *argv[])
...
+    clock_id = CLOCK_REALTIME;
 
     // Arg-parsing is only supported on Linux.
     int rv = parse_arguments(argc, argv, &clock_id);
     if (rv) {
         errExit("Argument error.");
     }
 
-    svr_clock_getres();
+#if !defined(L3_ENABLED)
```
gregthelaw, line 188: Is this right? Why only call `svr_clock_calibrate()` if there is no logging enabled? It looks like this artificially reduces the reported performance when logging is enabled.

gregthelaw:
OK, things seem to make a bit more sense if I stub out the clock calibration stuff.
```
gel@onyx:~/src/l3$ git diff
diff --git a/use-cases/client-server-msgs-perf/svmsg_file_server.c b/use-cases/client-server-msgs-perf/svmsg_file_server.c
index 65b146f..483df0f 100644
--- a/use-cases/client-server-msgs-perf/svmsg_file_server.c
+++ b/use-cases/client-server-msgs-perf/svmsg_file_server.c
@@ -185,9 +185,7 @@ main(int argc, char *argv[])
         errExit("Argument error.");
     }
 
-#if !defined(L3_ENABLED)
-    svr_clock_calibrate();
-#endif  // L3_ENABLED
+    // svr_clock_calibrate();
 
     /* Create server message queue */
 
gel@onyx:~/src/l3$ ./test.sh run-all-client-server-perf-tests --clock-default $((1000 * 1000)) > ~/tmp/run-all-perf.client-svr-throughput.1Mil.5ct.out 2>&1
gel@onyx:~/src/l3$ egrep -E 'Server throughput' ~/tmp/run-all-perf.client-svr-throughput.1Mil.5ct.out| cut -f2,6,7 -d','
 No logging, Server throughput=499212 (~499.21 K) ops/sec, Client throughput=202870 (~202.87 K) ops/sec
 L3-logging (no LOC), Server throughput=495074 (~495.07 K) ops/sec, Client throughput=197215 (~197.21 K) ops/sec
 L3-fast logging (no LOC), Server throughput=483375 (~483.37 K) ops/sec, Client throughput=189332 (~189.33 K) ops/sec
 L3-logging default LOC, Server throughput=492963 (~492.96 K) ops/sec, Client throughput=196017 (~196.01 K) ops/sec
 L3-logging LOC-ELF, Server throughput=494008 (~494.00 K) ops/sec, Client throughput=197329 (~197.32 K) ops/sec
```
so it's showing less than 1% slowdown. Ignore the l3-fast result, I think I know why the C version is sometimes faster, I have a fix for the assembly version, I'll create a PR for it next week.

use-cases/client-server-msgs-perf/svmsg_file_server.c
gapisback: Think of this as a diagnostic code now. I added this to understand what the impact of a call to `clock_gettime()` is.

Previously, I played around with ways to deduct this overhead from the overall elapsed time computation, to see if it makes a difference. It did not, so I don't really use the calibrated overhead value anywhere in metrics computation.

Also, note this call occurs on L188 whereas the actual `start` begins much later in the code:

```
260     if (clock_gettime(clock_id, &ts0)) {    // ***** Timing begins
261         errExit("clock_gettime-ts0");
262     }
```

So, your comment below is also moot and this `calibrate` call is a red herring.

gapisback:
@gregthelaw : I think in your results when you turned calibration ON v/s OFF:

- Calibration ON:

```
gel@onyx:~/src/l3$ egrep -E 'Server throughput' ~/tmp/run-all-perf.client-svr-throughput.1Mil.5ct.out| cut -f2,6,7 -d','
No logging, Server throughput=512189 (~512.18 K) ops/sec, Client throughput=186824 (~186.82 K) ops/sec
 L3-logging (no LOC), Server throughput=480790 (~480.79 K) ops/sec, Client throughput=186319 (~186.31 K) ops/sec
```
Server throughput: 513 -> 481 : Drop of 6%
Client  throughput: 186 -> 186 : Drop of ~0%

- Calibration OFF:

```
gel@onyx:~/src/l3$ egrep -E 'Server throughput' ~/tmp/run-all-perf.client-svr-throughput.1Mil.5ct.out| cut -f2,6,7 -d','
 No logging, Server throughput=499212 (~499.21 K) ops/sec, Client throughput=202870 (~202.87 K) ops/sec
 L3-logging (no LOC), Server throughput=495074 (~495.07 K) ops/sec, Client throughput=197215 (~197.21 K) ops/sec
```

Server throughput: 499 -> 495 : Drop of <1%
Client  throughput: 202 -> 197 : Drop of ~2.5%

----

I don't think this is explained by turning OFF calibration.

That call happens for baseline but there is also a `sleep 5` in `test.sh` between server and client start:

```
604 function build-and-run-client-server-perf-test()
[...]
657     ${server_bin} ${SvrClockArg} &
658
659     set +x
660     sleep 5
661
662     local numclients=${NumClients}
663     local ictr=0
664     # shellcheck disable=SC2086
665     while [ ${ictr} -lt ${numclients} ]; do
666
667         set -x
668         ${client_bin} "${num_msgs_per_client}" &
```


All server and client timing begins -after- calibration is completed.

The 5-second wait should be sufficient for calibration to complete, and for server to get into wait-for() loop before any clients fire up.

To address your concern:

- I would still like to retain the call to calibration so that when anyone else runs these scripts on some other machines, we get a quick read into how the clocks are performing on their system.
- I will move this call to the end of the server processing after all workload is run, which will take away any doubt that calibration for baseline is affecting metrics.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -293,40 +306,28 @@ main(int argc, char *argv[])
...
-
-            if (clock_gettime(clock_id, &ts1)) {    // Cumulative timing ends
-                errExit("clock_gettime-ts1");
-            }
-            nsec1 = timespec_to_ns(&ts1);
-            elapsed_ns = (nsec1 - nsec0);
-            clientp->cumu_time_ns += elapsed_ns;    // Cumulative elapsed-ns
+            break;
 
+          case REQ_MT_SET_THROUGHPUT:
```
gapisback, line 326: New op-code sent by client to register its client-side throughput metric with the server.

Server can then aggregate avg. client-side throughput across all clients, in its summary report.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -340,54 +341,68 @@ main(int argc, char *argv[])
...
                    throughput, value_str(throughput),
                    NumActiveClients);
 
-            if (NumActiveClients == 0) {
-                break;
-            }
             // Client has exited, so no need to send ack-response back to it.
             req.clientId = 0;
+            if (NumActiveClients == 0) {
+                goto end_forever_loop;
```
gapisback, line 363: This fixes a bug in previous version where, under some cases, we would be [incorrectly] doing a `break` here. This trips up a latter assertion checking for `#-of-clients-left == 0`.

I've seen this assertion trip-up in my own dev-runs, and this fix has remedied that problem.

CI-jobs that were previously failing (in `/main` also) seems to be succeeding well now.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -340,54 +341,68 @@ main(int argc, char *argv[])
...
            NumActiveClients, NumActiveClientsHWM);
 
-    printSummaryStats(ActiveClients, NumActiveClientsHWM, clock_id);
+    printSummaryStats(run_descr, ActiveClients, NumActiveClientsHWM, clock_id,
+                      elapsed_ns);
+
+    // For visibility into how clocks are performing on user's machine,
+    // run clock-calibration after all workload / metrics collection is done.
+#if !defined(L3_ENABLED)
+    svr_clock_calibrate();
```
gapisback, line 404: Relocate here to avoid perturbing CPU-clocking and perf-metrics while workload is running.

gregthelaw (approved):

PR #63, opened 2024-05-28
[Perf] Support L3_LOGT_FPRINTF -> l3_log_fprintf() for benchmarking.

gapisback:
This commit adds support for `L3_LOGT_FPRINTF=1 make client-server-perf-test` to build client-server perf-tests to log using the `fprintf()` system call.

Makefile and #if'def-ed code is added to support fprintf() logging. `l3_log_fprintf()` defined as static inline to avoid fn-call overhead. 

test.sh - Add "fprintf" support to existing test-method.

Fix seg-fault running fprintf() logging in debug mode.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,7 @@ help::
 	@echo ' '
 	@echo 'To build client-server performance test programs and run performance test(s)'
 	@echo ' make clean && CC=gcc LD=g++ L3_ENABLED=0         make client-server-perf-test  # Baseline'
+	@echo ' make clean && CC=gcc LD=g++ L3_LOGT_FPRINTF=1    make client-server-perf-test  # fprintf() logging'
 	@echo ' make clean && CC=gcc LD=g++                      make client-server-perf-test  # L3-logging'
 	@echo ' make clean && CC=gcc LD=g++ L3_FASTLOG_ENABLED=1 make client-server-perf-test  # L3 Fast logging'
 	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=1     make client-server-perf-test  # L3+LOC logging'
@@ -572,11 +573,17 @@ ifeq ($(L3_ENABLED), $(L3_DEFAULT))
 
 # Under L3-logging, invoke the L3-fast logging API, which needs the .S file
 ifeq ($(L3_FASTLOG_ENABLED), 1)
+
     CLIENT_SERVER_NON_MAIN_SRCS += $(L3_ASSEMBLY)
     CFLAGS += -DL3_FASTLOG_ENABLED
-endif
+
+else ifeq ($(L3_LOGT_FPRINTF), 1)
+
+    CFLAGS += -DL3_LOGT_FPRINTF
 
 endif
+
+endif   # L3_ENABLED
 
 # Name of LOC-generated source file for the client-server perf-test program.
 ifeq ($(L3_LOC_ENABLED), $(L3_LOC_DEFAULT))
--- a/include/l3.h
+++ b/include/l3.h
@@ -55,11 +55,15 @@ typedef enum {
       L3_LOG_UNDEF      = 0
     , L3_LOG_MMAP
     , L3_LOG_FPRINTF
+    , L3_LOGTYPE_MAX
     , L3_LOG_DEFAULT    = L3_LOG_MMAP
 } l3_log_t;
 
+extern FILE *  l3_log_fh;       // L3_LOG_FPRINTF: Opened by fopen()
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -572,12 +573,18 @@ ifeq ($(L3_ENABLED), $(L3_DEFAULT))
...
 # Under L3-logging, invoke the L3-fast logging API, which needs the .S file
 ifeq ($(L3_FASTLOG_ENABLED), 1)
+
     CLIENT_SERVER_NON_MAIN_SRCS += $(L3_ASSEMBLY)
     CFLAGS += -DL3_FASTLOG_ENABLED
-endif
+
+else ifeq ($(L3_LOGT_FPRINTF), 1)
+
+    CFLAGS += -DL3_LOGT_FPRINTF
```
gapisback, line 582: Env-var `L3_LOGT_FPRINTF` translated down to compile code conditionally defined under `-DL3_LOGT_FPRINTF`.

test.sh
```diff
@@ -522,6 +522,19 @@ function test-build-and-run-client-server-perf-test()
...
 
     set +x
+
+    echo " "
+    echo "**** ${Me}: Client-server performance testing with L3-fprintf logging ON:"
+    echo " "
+    build-and-run-client-server-perf-test "${num_msgs_per_client}"  \
+                                          "${l3_log_enabled}"       \
+                                          "${l3_LOC_disabled}"      \
+                                          "fprintf"
```
gapisback, line 532: Previous refactoring that introduced `build-and-run-client-server-perf-test()` test-method now nicely serves us to add `"fprintf"` directive, to drive fprintf()-logging.

src/l3.c
```diff
@@ -242,6 +271,28 @@ l3_init(const char *path)
...
+{
+    if (!path) {
+        fprintf(stderr, "%s: Invalid arg 'path'=%p\n", __func__, path);
+        return -1;
+    }
+
+    l3_log_fh = fopen(path, "w+");
+    if (!l3_log_fh) {
+        fprintf(stderr, "%s: Error opening log-file at '%p'. Error=%d\n",
+                __func__, path, errno);
```
gregthelaw, line 289: Minor: you could use `strerror()` to give a more meaningful error message. Or come to that, `perror()`

gregthelaw (approved):
Looks good; one minor suggestion, feel free to ignore.

include/l3.h
```diff
@@ -143,6 +162,15 @@ void l3_log_mmap(const char *msg, const uint64_t arg1, const uint64_t arg2,
                  const uint32_t loc);
 #endif  // L3_LOC_ENABLED
 
+/**
+ * l3_log_fprintf() - Log a msg to a file using fprintf().
+ */
+static inline void
+l3_log_fprintf(const char *msg, const uint64_t arg1, const uint64_t arg2)
+{
+    fprintf(l3_log_fh, msg, arg1, arg2);
```
gregthelaw, line 171: Just a note that this is fine for benchmarking purposes, but won't work on all architectures (namely i386) if the strings format specifiers aren't 64-bit (e.g. `%c` or `%d`).

src/l3.c
gapisback: Sounded like a good idea, so I added this.

Upon unit-testing, I discovered that the existing error raising mechanism in this core programs are already reporting what seems to be like an `perror` output:

See this output, when I changed an output file to `const char *logfile = "/xxtmp/l3.c-server-test.dat";`

```
agurajada-Linux-Vm:[939] $ build/release/bin/use-cases/svmsg_file_server
l3_init_fprintf: Error opening log-file at '0x5de2ee48760b'. Error=2
fopen failed: No such file or directory
ERROR [ENOENT No such file or directory] l3_log_init
```

So, I think a specific addition of `perror()` in this one place is not needed.

While investigating this, it occurred to me that we should enhance this errExit() and similar interfaces to integrate with LOC. I have opened a new dev-issue #68 to track this potential enhancement.

include/l3.h
gapisback: The idea, and expectation, here is that the printf() format-specifier validation I added (in debug mode, under SHA f280675) is supposed to come into play.

If the specific platform requires a different format-specifiers, code author is supposed to invoke it with those, othewise, this will fail in debug builds, defined in `l3.h`:

```
 89             } else if (0) {                                             \
 90                 printf((msg), (arg1), (arg2));                          \
```

I agree that now as the `fprintf()` interfaces are going public, we have to do a minor revisit on usability. Let's hold-off on that till we get these changes in first.

PR #64, opened 2024-05-28
[Perf] Support L3_LOGT_WRITE -> l3_log_write() for benchmarking.

gapisback:
This commit adds support for `L3_LOGT_WRITE=1 make client-server-perf-test` to build client-server perf-tests to log using write() system call.

- Makefile and #if'def-ed code is added to support write() logging.
- test.sh - Add "write" logging-support to existing test-method.

-----

###   [Perf] Add fprintf/write-logging perf unit-tests

This commit adds couple of new unit-tests to do standalone micro-benchmarking of the performance of l3_log_fprintf(),
 vs l3_log_write() interfaces.

- `make run-unit-tests` will automatically invoke these new tests.
- To resolve usefulness of `write()` v/s writing out a formatted a message, this perf unit-test provides an alternate implementation of `write_msg()` interface thru `l3_log()`.
 
Details in the descriptions below.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,8 +51,7 @@ help::
 	@echo 'To build client-server performance test programs and run performance test(s)'
 	@echo ' make clean && CC=gcc LD=g++ L3_ENABLED=0         make client-server-perf-test  # Baseline'
 	@echo ' make clean && CC=gcc LD=g++ L3_LOGT_FPRINTF=1    make client-server-perf-test  # fprintf() logging'
-	@echo ' make clean && CC=gcc LD=g++ L3_LOGT_WRITE=1      make client-server-perf-test  # write() logging'
-	@echo ' make clean && CC=gcc LD=g++ L3_LOGT_WRITE=2      make client-server-perf-test  # write()-msg logging'
+	@echo ' make clean && CC=gcc LD=g++ L3_LOGT_WRITE=1      make client-server-perf-test  # write() msg logging'
 	@echo ' make clean && CC=gcc LD=g++                      make client-server-perf-test  # L3-logging'
 	@echo ' make clean && CC=gcc LD=g++ L3_FASTLOG_ENABLED=1 make client-server-perf-test  # L3 Fast logging'
 	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=1     make client-server-perf-test  # L3+LOC logging'
@@ -603,10 +602,6 @@ else ifeq ($(L3_LOGT_FPRINTF), 1)
 else ifeq ($(L3_LOGT_WRITE), 1)
 
     CFLAGS += -DL3_LOGT_WRITE
-
-else ifeq ($(L3_LOGT_WRITE), 2)
-
-    CFLAGS += -DL3_LOGT_WRITE_MSG
 
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -51,10 +51,12 @@ help::
 	@echo 'To build client-server performance test programs and run performance test(s)'
 	@echo ' make clean && CC=gcc LD=g++ L3_ENABLED=0         make client-server-perf-test  # Baseline'
 	@echo ' make clean && CC=gcc LD=g++ L3_LOGT_FPRINTF=1    make client-server-perf-test  # fprintf() logging'
+	@echo ' make clean && CC=gcc LD=g++ L3_LOGT_WRITE=1      make client-server-perf-test  # write() logging'
 	@echo ' make clean && CC=gcc LD=g++                      make client-server-perf-test  # L3-logging'
 	@echo ' make clean && CC=gcc LD=g++ L3_FASTLOG_ENABLED=1 make client-server-perf-test  # L3 Fast logging'
 	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=1     make client-server-perf-test  # L3+LOC logging'
 	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=2     make client-server-perf-test  # L3+LOC-ELF logging'
+
 	@echo ' '
 	@echo 'To build L3-sample programs with LOC-enabled and run unit-tests:'
 	@echo ' make clean && CC=gcc LD=g++         L3_LOC_ENABLED=1 make all-c-tests   && L3_LOC_ENABLED=1 make run-c-tests'
@@ -580,6 +582,10 @@ ifeq ($(L3_FASTLOG_ENABLED), 1)
 else ifeq ($(L3_LOGT_FPRINTF), 1)
 
     CFLAGS += -DL3_LOGT_FPRINTF
+
+else ifeq ($(L3_LOGT_WRITE), 1)
+
+    CFLAGS += -DL3_LOGT_WRITE
 
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -419,6 +419,10 @@ L3_C_UNIT_TEST_BIN  := $(BINDIR)/$(UNIT_DIR)/l3_dump.py-test
 LOC_MACRO_TEST_BIN  := $(LOC_MACRO_TEST_CPP_PROGRAM_BIN)
 SIZE_UNIT_TEST_BIN  := $(BINDIR)/$(UNIT_DIR)/size_str-test
 
+# L3-logging interfaces' performance unit-tests
+FPRINTF_PERF_UNIT_TEST_BIN  := $(BINDIR)/$(UNIT_DIR)/l3-fprintf-perf-test
+WRITE_PERF_UNIT_TEST_BIN    := $(BINDIR)/$(UNIT_DIR)/l3-write-perf-test
+
 # ##############################################################################
 # Generate symbols and dependencies to build unit-test sources
 # ##############################################################################
@@ -463,6 +467,13 @@ else
 
 $(BINDIR)/$(UNIT_DIR)/l3_dump.py-test: $(OBJDIR)/$(UNITTESTS_DIR)/l3_dump.py-test.o \
                                         $(OBJDIR)/$(SRCDIR)/l3.o
+
+$(BINDIR)/$(UNIT_DIR)/l3-fprintf-perf-test: $(OBJDIR)/$(UNITTESTS_DIR)/l3-fprintf-perf-test.o \
+                                            $(OBJDIR)/$(SRCDIR)/l3.o
+
+$(BINDIR)/$(UNIT_DIR)/l3-write-perf-test: $(OBJDIR)/$(UNITTESTS_DIR)/l3-write-perf-test.o \
+                                          $(OBJDIR)/$(SRCDIR)/l3.o
+
 endif
 
 # We only need this extra include to find size_str.h, needed to build the
@@ -470,6 +481,11 @@ endif
 $(BINDIR)/$(UNIT_DIR)/size_str-test: INCLUDE += -I ./$(L3_UTILSDIR)
 $(BINDIR)/$(UNIT_DIR)/size_str-test: $(OBJDIR)/$(UNITTESTS_DIR)/size_str-test.o \
                                      $(OBJDIR)/$(L3_UTILSDIR)/size_str.o
+
+# Unit-test for l3-write interfaces needs this defined so that l3_log()
+# gets correctly redefined to l3_log_fprintf() / l3_log_write() interface
+$(BINDIR)/$(UNIT_DIR)/l3-fprintf-perf-test: DFLAGS_UNIT := -DL3_LOGT_FPRINTF
+$(BINDIR)/$(UNIT_DIR)/l3-write-perf-test: DFLAGS_UNIT := -DL3_LOGT_WRITE
 
 # ###################################################################
 # Report build machine details and compiler version for troubleshooting,
@@ -729,7 +745,7 @@ $(BINDIR)/%/.:
 #
 # For all-test-code, we need to use -I test-code/<subdir>
 # Dependencies for the main executables
-COMPILE.c       = $(CC)  -x c $(CFLAGS) $(INCLUDE) -c
+COMPILE.c       = $(CC)  -x c $(CFLAGS) $(DFLAGS_UNIT) $(INCLUDE) -c
 COMPILE.cpp     = $(CXX) -x c++ $(CPPFLAGS) $(CFLAGS) $(INCLUDE) -c
 COMPILE.cc      = $(CXX) -x c++ $(CFLAGS) $(INCLUDE) -c
 COMPILE.loc.c   = $(LOC_C_CC) $(CFLAGS) $(INCLUDE) -c
@@ -817,6 +833,13 @@ run-unit-tests: all-unit-tests
 	python3 $(L3_DUMP) $(L3_DUMP_ARG_LOG_FILE) $(L3_C_UNIT_FAST_LOG_TEST_DATA) $(L3_DUMP_ARG_BINARY) ./$(L3_C_UNIT_TEST_BIN)
 	@echo
 	./$(SIZE_UNIT_TEST_BIN)
+	@echo
+	./$(FPRINTF_PERF_UNIT_TEST_BIN)
+	# L3-write performance test seems to work better on subsequent runs.
+	@echo
+	./$(WRITE_PERF_UNIT_TEST_BIN)
+	@echo
+	./$(WRITE_PERF_UNIT_TEST_BIN)
 
 run-loc-tests: all-loc-tests
 	@echo
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,7 @@ help::
 	@echo ' make clean && CC=gcc LD=g++ L3_ENABLED=0         make client-server-perf-test  # Baseline'
 	@echo ' make clean && CC=gcc LD=g++ L3_LOGT_FPRINTF=1    make client-server-perf-test  # fprintf() logging'
 	@echo ' make clean && CC=gcc LD=g++ L3_LOGT_WRITE=1      make client-server-perf-test  # write() logging'
+	@echo ' make clean && CC=gcc LD=g++ L3_LOGT_WRITE=2      make client-server-perf-test  # write()-msg logging'
 	@echo ' make clean && CC=gcc LD=g++                      make client-server-perf-test  # L3-logging'
 	@echo ' make clean && CC=gcc LD=g++ L3_FASTLOG_ENABLED=1 make client-server-perf-test  # L3 Fast logging'
 	@echo ' make clean && CC=gcc LD=g++ L3_LOC_ENABLED=1     make client-server-perf-test  # L3+LOC logging'
@@ -602,6 +603,10 @@ else ifeq ($(L3_LOGT_FPRINTF), 1)
 else ifeq ($(L3_LOGT_WRITE), 1)
 
     CFLAGS += -DL3_LOGT_WRITE
+
+else ifeq ($(L3_LOGT_WRITE), 2)
+
+    CFLAGS += -DL3_LOGT_WRITE_MSG
 
 endif
 
--- a/src/l3.c
+++ b/src/l3.c
@@ -379,34 +379,11 @@ l3_log_mmap(const char *msg, const uint64_t arg1, const uint64_t arg2,
 
 /**
  * l3_log_write() - 'C' interface to log L3 log-entries using write()
+ * The user's msg is sprintf()'ed using 'msgfmt' format specifiers, requiring
+ * 2 arguments.
  */
 void
-l3_log_write(const char *msg, const uint64_t arg1, const uint64_t arg2)
-{
-    L3_ENTRY le     = {0};
-    L3_ENTRY *lep   = &le;
-
-    lep->tid = l3_mytid();
-
-    lep->msg = msg;
-    lep->arg1 = arg1;
-    lep->arg2 = arg2;
-
-    if (write(l3_log_fd, lep, sizeof(*lep)) != sizeof(*lep)) {
-        fprintf(stderr, "%s(): write() failed. errno=%d\n",
-                __func__, errno);
-        return;
-    }
-    return;
-}
-
-/**
- * l3_log_write_msg() - 'C' interface to log L3 log-entries using write()
- * Similar to l3_log_write(), but the msg is sprintf()'ed using 'msgfmt' format
- * specifiers, requiring 2 arguments.
- */
-void
-l3_log_write_msg(const char *msgfmt, const uint64_t arg1, const uint64_t arg2)
+l3_log_write(const char *msgfmt, const uint64_t arg1, const uint64_t arg2)
 {
     char msgbuf[255];
 
--- a/src/l3.c
+++ b/src/l3.c
@@ -122,6 +122,8 @@ typedef struct l3_log
 
 #define L3_ARRAY_LEN(arr)   (sizeof(arr) / sizeof(*arr))
 
+#define L3_MIN(a, b)        ((a) > (b) ? (b) : (a))
+
 /**
  * Support for different 'logging' types under L3's APIs.
  */
@@ -129,6 +131,7 @@ static const char * L3_logtype_name[] = {
                           "L3_LOG_unknown"
                         , "L3_LOG_MMAP"
                         , "L3_LOG_FPRINTF"
+                        , "L3_LOG_WRITE"
                 };
 
 L3_STATIC_ASSERT((L3_ARRAY_LEN(L3_logtype_name) == L3_LOGTYPE_MAX),
@@ -138,7 +141,9 @@ L3_STATIC_ASSERT((L3_ARRAY_LEN(L3_logtype_name) == L3_LOGTYPE_MAX),
 L3_LOG *l3_log = NULL;      // L3_LOG_MMAP: Also referenced in l3.S for
                             // fast-logging.
 
-FILE *  l3_log_fh = NULL;  // L3_LOG_FPRINTF: Opened by fopen()
+FILE *  l3_log_fh = NULL;   // L3_LOG_FPRINTF: Opened by fopen()
+
+int     l3_log_fd = -1;     // L3_LOG_WRITE: Opened by open()
 
 /**
  * The L3-dump script expects a specific layout and its parsing routines
@@ -155,6 +160,7 @@ L3_STATIC_ASSERT(offsetof(L3_LOG,slots) == sizeof(L3_ENTRY),
  * ****************************************************************************
  */
 int l3_init_fprintf(const char *path);
+int l3_init_write(const char *path);
 
 
 #if __APPLE__
@@ -193,6 +199,10 @@ l3_log_init(const l3_log_t logtype, const char *path)
 
       case L3_LOG_FPRINTF:
         rv = l3_init_fprintf(path);
+        break;
+
+      case L3_LOG_WRITE:
+        rv = l3_init_write(path);
         break;
 
       default:
@@ -294,6 +304,28 @@ l3_init_fprintf(const char *path)
     return 0;
 }
 
+/**
+ * ****************************************************************************
+ * Initialize L3's logging sub-system to use write() to named `path`.
+ */
+int
+l3_init_write(const char *path)
+{
+    if (!path) {
+        fprintf(stderr, "%s: Invalid arg 'path'=%p\n", __func__, path);
+        return -1;
+    }
+
+    l3_log_fd = open(path, (O_RDWR | O_CREAT), 0666);
+    if (l3_log_fd == -1) {
+        fprintf(stderr, "%s: Error opening log-file at '%p'. Error=%d\n",
+                __func__, path, errno);
+        return -1;
+    }
+    printf("Initialized write() logging to '%s'\n", path);
+    return 0;
+}
+
 // ****************************************************************************
 static pid_t
 l3_mytid(void)
@@ -341,6 +373,54 @@ l3_log_mmap(const char *msg, const uint64_t arg1, const uint64_t arg2,
     l3_log->slots[idx].msg = msg;
     l3_log->slots[idx].arg1 = arg1;
     l3_log->slots[idx].arg2 = arg2;
+}
+
+/**
+ * l3_log_write() - 'C' interface to log L3 log-entries using write()
+ */
+void
+l3_log_write(const char *msg, const uint64_t arg1, const uint64_t arg2)
+{
+    L3_ENTRY le     = {0};
+    L3_ENTRY *lep   = &le;
+
+    lep->tid = l3_mytid();
+
+    lep->msg = msg;
+    lep->arg1 = arg1;
+    lep->arg2 = arg2;
+
+    if (write(l3_log_fd, lep, sizeof(*lep)) != sizeof(*lep)) {
+        fprintf(stderr, "%s(): write() failed. errno=%d\n",
+                __func__, errno);
+        return;
+    }
+    return;
+}
+
+/**
+ * l3_log_write_msg() - 'C' interface to log L3 log-entries using write()
+ * Similar to l3_log_write(), but the msg is sprintf()'ed using 'msgfmt' format
+ * specifiers, requiring 2 arguments.
+ */
+void
+l3_log_write_msg(const char *msgfmt, const uint64_t arg1, const uint64_t arg2)
+{
+    char msgbuf[255];
+
+    // snprintf() returns #-bytes that _would_ have been written if the msgbuf
+    // was large enough.
+    int msglen = snprintf(msgbuf, sizeof(msgbuf), msgfmt, arg1, arg2);
+
+    // So, write-to-the-log only the actual data that is in the buffer.
+    int datalen = L3_MIN(msglen, sizeof(msgbuf));
+
+    if (write(l3_log_fd, msgbuf, datalen) != datalen) {
+        fprintf(stderr, "%s(): write() failed, datalen=%d. errno=%d\n",
+                __func__, datalen, errno);
+        return;
+    }
+    return;
 }
 
 /**
--- a/src/l3.c
+++ b/src/l3.c
@@ -132,6 +132,7 @@ static const char * L3_logtype_name[] = {
                         , "L3_LOG_MMAP"
                         , "L3_LOG_FPRINTF"
                         , "L3_LOG_WRITE"
+                        , "L3_LOG_WRITE_MSG"
                 };
 
 L3_STATIC_ASSERT((L3_ARRAY_LEN(L3_logtype_name) == L3_LOGTYPE_MAX),
@@ -202,6 +203,7 @@ l3_log_init(const l3_log_t logtype, const char *path)
         break;
 
       case L3_LOG_WRITE:
+      case L3_LOG_WRITE_MSG:
         rv = l3_init_write(path);
         break;
 
--- a/tests/unit/l3-fprintf-perf-test.c
+++ b/tests/unit/l3-fprintf-perf-test.c
@@ -1,0 +1,92 @@
+/**
+ * *****************************************************************************
+ * \file l3-fprintf-perf-test.c
+ * \author Aditya P. Gurajada
+ * \brief L3: Lightweight Logging Library fprintf() logging APIs Unit-test
+ * \version 0.1
+ * \date 2024-05-29
+ *
+ * \copyright Copyright (c) 2024
+ *
+ * Usage: program-name [ <number-of-million-msgs> ]
+ *  Default, run fprintf()-logging perf-tests on 1 million messages
+ * *****************************************************************************
+ */
+#define _POSIX_C_SOURCE 199309L
+
+#include <stdlib.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <string.h>
+
+#include "l3.h"
+#include "l3-perf-test.h"
+
+void test_fprintf_logging_perf(int nMil);
+
+/**
+ * *****************************************************************************
+ * The L3 logging-APIs to exercise logging using fprintf() system call are
+ * exercised in this unit-test. As a convenience, we just log a large # of
+ * entries to get a rough measure of logging throughput using this interface.
+ * *****************************************************************************
+ */
+int
+main(const int argc, const char * argv[])
+{
+    int nMil = 1;
+    if (argc > 1) {
+        nMil = atoi(argv[1]);
+    }
+
+    test_fprintf_logging_perf(nMil);
+
+    return 0;
+}
+
+#if L3_LOGT_FPRINTF
+
+void
+test_fprintf_logging_perf(int nMil)
+{
+    // Logging perf benchmarking methods below share this log-file.
+    const char *logfile = "/tmp/l3-fprintf-logging-test.dat";
+    int e = l3_log_init(L3_LOG_FPRINTF, logfile);
+    if (e) {
+        abort();
+    }
+    test_logging_perf("fprintf", nMil, logfile);
+}
+
+#endif  // L3_LOGT_FPRINTF
+
+/**
+ * *****************************************************************************
+ * test_logging_perf() - Common function to log n-msgs and measure perf metrics.
+ * NOTE: This cannot be a shared function in common .c file as the build-time
+ * -D<flag> dictates which version of l3_log() [ write / fprintf ] will be used.
+ * *****************************************************************************
+ */
+void
+test_logging_perf(const char *logtype, int nMil, const char *filename)
+{
+    struct timespec ts0;
+    struct timespec ts1;
+    if (clock_gettime(CLOCK_REALTIME, &ts0)) {
+        abort();
+    }
+
+    uint32_t n = 0;
+    for (n = 0; n < (nMil * L3_MILLION); n++) {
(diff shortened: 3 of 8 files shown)


gregthelaw (commented):
So this doesn't write the log message itself, but instead the ptr to the string literal and args, and requires post processing as for l3 logging, right?

I'm not sure how sensible this is. Why would you ever do things this way? As opposed to say using `write()` to just write the log message directly. I would expect the performance between the two techniques to be indistinguishable, but writing the string directly doesn't require post processing.

src/l3.c
```diff
@@ -342,6 +372,29 @@ l3_log_mmap(const char *msg, const uint64_t arg1, const uint64_t arg2,
...
+    L3_ENTRY le     = {0};
+    L3_ENTRY *lep   = &le;
+
+    lep->tid = l3_mytid();
+
+    lep->msg = msg;
+    lep->arg1 = arg1;
+    lep->arg2 = arg2;
+
+    if (write(l3_log_fd, lep, sizeof(*lep) != sizeof(*lep))) {
```
gregthelaw, line 390: @gapisback I think this line has a bug - the closing parens at the end are out of order, so it always gives `len` argument to `write()` of 0.

gregthelaw (changes requested):

gapisback:
@gregthelaw W.r.t [this comment](https://github.com/undoio/l3/pull/64#pullrequestreview-2085104391) of yours, here is what I reasoned:

As per [this Gnu reference](https://man7.org/linux/man-pages/man3/fwrite.3p.html) on `write()` or `fwrite()`, this writes out binary data:

```
The fwrite() function shall write, from the array pointed to by ptr, up to nitems 
elements whose size is specified by size, to the stream pointed to by stream.
```

fprintf() is useful for format-string processing. 

```
Each of these functions converts, formats, and prints its
       arguments under control of the format. 
```

So you are doing more work at run-time while logging using `fprintf()` v/s `fwrite()`.

----

Re: _As opposed to say using write() to just write the log message directly. I would expect the performance between the two techniques to be indistinguishable, but writing the string directly doesn't require post processing._

I don't think that's entirely true. In order to `write` out the file's name, you would have to pre-format an output buffer @ run-time and `write` that out. That means, L3-logging has to do the work up-front to convert the %s to the file's name etc.

----


Going by [this blog post](https://stackoverflow.com/questions/15079049/write-vs-fprintf-why-different-and-which-is-better), `write` does unbuffered writes to the file, as opposed to `fprintf` or `fwrite`, which may involve file-system buffering.

So, I thought supporting `write` is better (than `fwrite`, which may do file-system buffering) and also useful in that we can write out binary data to the file, just like it's being done with log to `mmap()`-ed file.

With default L3-logging, you are restricted to some n-log-entries. With `write` interface, theoretically the file can keep growing.

By benchmarking against `write`, I'm trying to compare the perf-gains of doing IO to mmap-'ed file, v/s the device IO thru `write` system call; with both schemes needing post-processing (so that part is the same).

Hope this helps.

src/l3.c
gapisback: Yes! You are right. Thanks, for catching it.

Will fix ... and re-measure overall throughput with write.

I'll see if I can whip-up a simple unit-test to exercise these new interfaces and verify that the file-size is coming out as expected.

If that's taking too much time, I'll punt and open an issue for a follow-through.

gapisback:
@gregthelaw : Fixing the bug in the `write` system call now shows a major difference in overall throughput.

With the bug, we were writing 0-bytes:

```
    **** Performance comparison for NumClients=5, NumOps=5000000 (5 Million) ****
+-------------------------------+-------------------+----------+-------------------+----------+
| Run-Type                      | Server throughput | Srv:Drop | Client throughput | Cli:Drop |
+-------------------------------+-------------------+----------+-------------------+----------+
| Baseline - No logging         |  ~99.19 K ops/sec |  0.00 %  |  ~22.86 K ops/sec |  0.00 %  |

| L3-write() logging (no LOC)   |  ~93.64 K ops/sec | -5.59 %  |  ~21.56 K ops/sec | -5.71 %  |
```

With the bug-fix, here's the new report:

```
    **** Performance comparison for NumClients=5, NumOps=5000000 (5 Million) ****
+-------------------------------+-------------------+----------+-------------------+----------+
| Run-Type                      | Server throughput | Srv:Drop | Client throughput | Cli:Drop |
+-------------------------------+-------------------+----------+-------------------+----------+
| Baseline - No logging         |  ~99.40 K ops/sec |  0.00 %  |  ~24.31 K ops/sec |  0.00 %  |

| L3-write() logging (no LOC)   |  ~77.63 K ops/sec | -21.90 % |  ~16.85 K ops/sec | -30.70 % |
```

Baseline is about the same in both runs ~99.19 K ops/sec.

With the bug-fix, the write-throughput drops from ~93.64 K to ~77.63 K ops/sec, resulting in -21.90 % over baseline.

This seems inline with expected performance of unbuffered write system call.

Do we need to dig more? (The bug-fix is warranted, certainly.)

gregthelaw:
> @gregthelaw W.r.t [this comment](https://github.com/undoio/l3/pull/64#pullrequestreview-2085104391) of yours, here is what I reasoned:
> 
> As per [this Gnu reference](https://man7.org/linux/man-pages/man3/fwrite.3p.html) on `write()` or `fwrite()`, this writes out binary data:
> 
> ```
> The fwrite() function shall write, from the array pointed to by ptr, up to nitems 
> elements whose size is specified by size, to the stream pointed to by stream.
> ```

It's true that `write()` can take can arbitrary binary data, but ASCII text is a subset of that. In fact, `printf()` will be implemented via a call to `write()`. Writing ASCII text like this is definitely a legit/supported thing to do.

> 
> fprintf() is useful for format-string processing.
> 
> ```
> Each of these functions converts, formats, and prints its
>        arguments under control of the format. 
> ```
> 
> So you are doing more work at run-time while logging using `fprintf()` v/s `fwrite()`.
> 
>> Re: _As opposed to say using write() to just write the log message directly. I would expect the performance between the two techniques to be indistinguishable, but writing the string directly doesn't require post processing._
> 
> I don't think that's entirely true. In order to `write` out the file's name, you would have to pre-format an output buffer @ run-time and `write` that out. That means, L3-logging has to do the work up-front to convert the %s to the file's name etc.

Sure, the the `write` is a syscall, and that is going to dominate. I'd be very surprised if the time taken to format were to be significant compare to the syscall overhead.

> 
> Going by [this blog post](https://stackoverflow.com/questions/15079049/write-vs-fprintf-why-different-and-which-is-better), `write` does unbuffered writes to the file, as opposed to `fprintf` or `fwrite`, which may involve file-system buffering.

Exactly. There are pros and cons, and it may be preferable to use `write()` vs `fprintf()`, but the time taken to format the string will be immaterial once you're doing a syscall, so there's no point doing the post-processing.

> 
> So, I thought supporting `write` is better (than `fwrite`, which may do file-system buffering) and also useful in that we can write out binary data to the file, just like it's being done with log to `mmap()`-ed file.
> 
> With default L3-logging, you are restricted to some n-log-entries. With `write` interface, theoretically the file can keep growing.
> 
> By benchmarking against `write`, I'm trying to compare the perf-gains of doing IO to mmap-'ed file, v/s the device IO thru `write` system call; with both schemes needing post-processing (so that part is the same).

Understood. I just don't see why anyone would want to do the `write()` with the post-processing. It won't be measurably faster and e.g. if you don't have your original binary handy the log will be unreadable. But I _do_ see why one won't do `write()` with formatted text.

src/l3.c
```diff
@@ -344,6 +374,49 @@ l3_log_mmap(const char *msg, const uint64_t arg1, const uint64_t arg2,
...
+    L3_ENTRY le     = {0};
+    L3_ENTRY *lep   = &le;
+
+    lep->tid = l3_mytid();
+
+    lep->msg = msg;
+    lep->arg1 = arg1;
+    lep->arg2 = arg2;
+
+    if (write(l3_log_fd, lep, sizeof(*lep)) != sizeof(*lep)) {
```
gapisback, line 391: @gregthelaw - ATT: bug-fix applied here and squashed into 1st commit.

Makefile
```diff
@@ -419,6 +419,10 @@ L3_C_UNIT_TEST_BIN  := $(BINDIR)/$(UNIT_DIR)/l3_dump.py-test
 LOC_MACRO_TEST_BIN  := $(LOC_MACRO_TEST_CPP_PROGRAM_BIN)
 SIZE_UNIT_TEST_BIN  := $(BINDIR)/$(UNIT_DIR)/size_str-test
 
+# L3-logging interfaces' performance unit-tests
+FPRINTF_PERF_UNIT_TEST_BIN  := $(BINDIR)/$(UNIT_DIR)/l3-fprintf-perf-test
+WRITE_PERF_UNIT_TEST_BIN    := $(BINDIR)/$(UNIT_DIR)/l3-write-perf-test
```
gapisback, line 424: We need two standalone API exercisers that also double-up as perf-unit tests.

The L3-logging calls are conditionally `#defined` via the `l3_log()` macro, so we need two build targets that will generate these two binaries.

Makefile
```diff
@@ -471,6 +482,11 @@ $(BINDIR)/$(UNIT_DIR)/size_str-test: INCLUDE += -I ./$(L3_UTILSDIR)
 $(BINDIR)/$(UNIT_DIR)/size_str-test: $(OBJDIR)/$(UNITTESTS_DIR)/size_str-test.o \
                                      $(OBJDIR)/$(L3_UTILSDIR)/size_str.o
 
+# Unit-test for l3-write interfaces needs this defined so that l3_log()
+# gets correctly redefined to l3_log_fprintf() / l3_log_write() interface
+$(BINDIR)/$(UNIT_DIR)/l3-fprintf-perf-test: DFLAGS_UNIT := -DL3_LOGT_FPRINTF
+$(BINDIR)/$(UNIT_DIR)/l3-write-perf-test: DFLAGS_UNIT := -DL3_LOGT_WRITE
```
gapisback, line 488: The -D flags needed for conditional compilation are added here, only for the appropriate `make` target.

tests/unit/l3-fprintf-perf-test.c
```diff
@@ -0,0 +1,92 @@
...
+
+    test_fprintf_logging_perf(nMil);
+
+    return 0;
+}
+
+#if L3_LOGT_FPRINTF
+
+void
+test_fprintf_logging_perf(int nMil)
```
gapisback, line 50: Define this conditionally, so that the build-time requirement of specifying `-DL3_LOGT_FPRINTF` is rigidly enforced.

src/l3.c
```diff
@@ -343,6 +373,48 @@ l3_log_mmap(const char *msg, const uint64_t arg1, const uint64_t arg2,
...
+    return;
+}
+
+/**
+ * l3_log_write_msg() - 'C' interface to log L3 log-entries using write()
+ * Similar to l3_log_write(), but the msg is sprintf()'ed using 'msgfmt' format
+ * specifiers, requiring 2 arguments.
+ */
+void
+l3_log_write_msg(const char *msgfmt, const uint64_t arg1, const uint64_t arg2)
```
gapisback, line 405: This is really a temp fn, added for perf-benchmarking - to figure out the overheads of a simple `write()` system call v/s the overhead of doing a `write()` on a msg-buffer where the print format specifiers have been replaced by the argument.

(This function should have gone into the 2nd commit, but got stuck in here. Eventually, this may come out altogether.)

tests/unit/l3-write-perf-test.c
```diff
@@ -0,0 +1,138 @@
...
+    const char *logfile = "/tmp/l3-write-logging-test.dat";
+    int e = l3_log_init(L3_LOG_WRITE, logfile);
+    if (e) {
+        abort();
+    }
+    test_logging_perf("write", nMil, logfile);
+}
+
+static void
+test_write_msg_logging_perf(int nMil)
```
gapisback, line 68: Variation test-method that will eventually invoke the `l3_log_write_msg()` interface.

gapisback:
@gregthelaw - To address your comment that we should invoke `write()` to log a fully-formatted message, I add unit-tests in this [2nd commit](https://github.com/undoio/l3/pull/64/commits/661a133eaf7d6f91d66125b761bcf87c9117ceb5).

The unit-tests serve multiple purposes:

1. `unit/l3-fprintf-perf-test` exercises L3-fprintf() interfaces added in previous commit. (Test-code coverage.)
2. `unit/l3-write-perf-test` does the same thing for l3-write() interfaces.
3. The `l3-write-perf-test` test exercises both variations of `write` call, to measure if there is any overhead of doing `snprintf()` to create a fully-formed message.

gapisback:
With these perf-unit test changes, here is some output on my Linux-VM, to log 100 million Msgs:

- fprintf():

```
agurajada-Linux-Vm:[1076] $ ./build/release/bin/unit/l3-fprintf-perf-test 100
Initialized fprintf() logging to '/tmp/l3-fprintf-logging-test.dat'
100 Mil fprintf() log msgs: 133 ns/msg (avg): /tmp/l3-fprintf-logging-test.dat

agurajada-Linux-Vm:[1077] $ ./build/release/bin/unit/l3-fprintf-perf-test 100
Initialized fprintf() logging to '/tmp/l3-fprintf-logging-test.dat'
100 Mil fprintf() log msgs: 124 ns/msg (avg): /tmp/l3-fprintf-logging-test.dat
```

There is no variation across multiple runs.

- write() v/s write_mg() - Logs the full message as a string:

Each invocation will log n-million messages thru both interfaces.

I found that the 1st run is always slow, so you need to run this workload multiple times for the metric to stabilize.
In the program, I also do a "warm-up" run, so ignore the 1st output below, for 1Mil msgs.
```
agurajada-Linux-Vm:[1078] $ ./build/release/bin/unit/l3-write-perf-test 100
Initialized write() logging to '/tmp/l3-write-logging-test.dat'
1 Mil write() log msgs: 390 ns/msg (avg): /tmp/l3-write-logging-test.dat

Initialized write() logging to '/tmp/l3-write-logging-test.dat'
100 Mil write() log msgs: 921 ns/msg (avg): /tmp/l3-write-logging-test.dat

Initialized write() logging to '/tmp/l3-writemsg-logging-test.dat'
100 Mil write() log msgs: 1127 ns/msg (avg): /tmp/l3-writemsg-logging-test.dat
```

write_msg() drop in performance is ~22%

- Subsequent runs of write perf unit-test:

```
agurajada-Linux-Vm:[1079] $ ./build/release/bin/unit/l3-write-perf-test 100                  # 100 million messages clocked
Initialized write() logging to '/tmp/l3-write-logging-test.dat'
1 Mil write() log msgs: 400 ns/msg (avg): /tmp/l3-write-logging-test.dat

Initialized write() logging to '/tmp/l3-write-logging-test.dat'
100 Mil write() log msgs: 435 ns/msg (avg): /tmp/l3-write-logging-test.dat

Initialized write() logging to '/tmp/l3-writemsg-logging-test.dat'
100 Mil write() log msgs: 547 ns/msg (avg): /tmp/l3-writemsg-logging-test.dat

agurajada-Linux-Vm:[1080] $ ./build/release/bin/unit/l3-write-perf-test 100
Initialized write() logging to '/tmp/l3-write-logging-test.dat'
1 Mil write() log msgs: 440 ns/msg (avg): /tmp/l3-write-logging-test.dat

Initialized write() logging to '/tmp/l3-write-logging-test.dat'
100 Mil write() log msgs: 427 ns/msg (avg): /tmp/l3-write-logging-test.dat

Initialized write() logging to '/tmp/l3-writemsg-logging-test.dat'
100 Mil write() log msgs: 534 ns/msg (avg): /tmp/l3-writemsg-logging-test.dat
```

- write(): Avg elapsed time across 2 runs: 431 ns/msg
- write_msg(): Avg elapsed time across 2 runs: 540 ns/msg == ~25% slower

gapisback:
Re: "_I just don't see why anyone would want to do the write() with the post-processing. It won't be measurably faster and ... _"

Unless I am doing something very wrong in my unit perf-test programs, we are seeing anywhere from 20-30% drop on my Linux-VM for write_msg() v/s write() interface.

And the bit about _e.g. if you don't have your original binary handy the log will be unreadable._ ... applies to default l3-mmap()-logging as well. So, I don't see that as a big objection.

gapisback:
@gregthelaw : Final comment on this item and a request from you. From my [previous comment](https://github.com/undoio/l3/pull/64#issuecomment-2137911488), the line below:

_Fixing the bug in the write system call now shows a major difference in overall throughput.
 With the bug-fix, the write-throughput drops from ~93.64 K to ~77.63 K ops/sec, resulting in -21.90 % over baseline._

Needs a re-validation from your side on your Linux-workstation. I re-ran the entire suite on my Linux-VM and do not see a big drop this time (after all changes, including the bug-fix are integrated:

Baseline:
```
| Run-Type                      | Server throughput | Srv:Drop | Client throughput | Cli:Drop |
+-------------------------------+-------------------+----------+-------------------+----------+
| Baseline - No logging         |  ~99.19 K ops/sec |  0.00 %  |  ~22.86 K ops/sec |  0.00 %  |
| L3-logging (no LOC)           |  ~99.60 K ops/sec |  0.41 %  |  ~23.44 K ops/sec |  2.51 %  |
| L3-write() logging (no LOC)   |  ~93.64 K ops/sec | -5.59 %  |  ~21.56 K ops/sec | -5.71 %  |
```

With fix:

```
    **** Performance comparison for NumClients=5, NumOps=5000000 (5 Million) ****
+-------------------------------+-------------------+----------+-------------------+----------+
| Run-Type                      | Server throughput | Srv:Drop | Client throughput | Cli:Drop |
+-------------------------------+-------------------+----------+-------------------+----------+
| Baseline - No logging         |  ~96.01 K ops/sec |  0.00 %  |  ~21.68 K ops/sec |  0.00 %  |
| L3-logging (no LOC)           | ~102.11 K ops/sec |  6.35 %  |  ~24.32 K ops/sec | 12.17 %  |
| L3-fast logging (no LOC)      |  ~88.77 K ops/sec | -7.54 %  |  ~19.75 K ops/sec | -8.88 %  |

| L3-write() logging (no LOC)   |  ~88.51 K ops/sec | -7.81 %  |  ~19.85 K ops/sec | -8.42 %  |
```

Now I see ~8% drop with write() system call, over the run w/o this bug-fix.

Can you re-run on your end and let's measure what the real overhead with simple `write()` call is for this workload.

src/l3.c
```diff
@@ -344,6 +374,49 @@ l3_log_mmap(const char *msg, const uint64_t arg1, const uint64_t arg2,
...
+{
+    char msgbuf[255];
+    int msglen = snprintf(msgbuf, sizeof(msgbuf), msgfmt, arg1, arg2);
+
+    if (write(l3_log_fd, msgbuf, msglen) != msglen) {
```
gregthelaw, line 410: I think you need `max(msglen, sizeof msgbuf)` as `snprintf` returns the number of bytes that would have been written -- i.e. can be more than `sizeof(msgbuf).

gapisback: I don't think so, @gregthelaw . The whole purpose of `snprintf()` is to ensure that there are no buffer overflows.

This interface is guaranteed to write max of (n - 1) bytes, with a null-terminator at the nth byte if formatted message's size exceeds buffer size.

Here's what the [printf man page](https://man7.org/linux/man-pages/man3/fprintf.3p.html) says:

```
 If n is zero, nothing shall be written and s may be a null pointer. 
Otherwise, output bytes beyond the n‐1st shall be discarded instead of being written to the array, 
and a null byte is written at the end of the bytes actually written into the array.
```

gregthelaw:
This is what I see on the `perf-l3_write` branch:
```
    **** Performance comparison for NumClients=5, NumOps=5000000 (5 Million) ****
+-------------------------------+-------------------+----------+-------------------+----------+
| Run-Type                      | Server throughput | Srv:Drop | Client throughput | Cli:Drop |
+-------------------------------+-------------------+----------+-------------------+----------+
| Baseline - No logging         | ~313.73 K ops/sec |  0.00 %  |  ~91.75 K ops/sec |  0.00 %  |
| L3-logging (no LOC)           | ~308.97 K ops/sec | -1.52 %  |  ~89.88 K ops/sec | -2.04 %  |
| L3-fast logging (no LOC)      | ~312.81 K ops/sec | -0.29 %  |  ~91.60 K ops/sec | -0.17 %  |
| L3-fprintf() logging (no LOC) | ~299.84 K ops/sec | -4.43 %  |  ~85.82 K ops/sec | -6.46 %  |
| L3-write() logging (no LOC)   | ~197.15 K ops/sec | -37.16 % |  ~49.13 K ops/sec | -46.46 % |
| L3-logging default LOC        | ~314.68 K ops/sec |  0.30 %  |  ~92.17 K ops/sec |  0.45 %  |
| L3-logging LOC-ELF            | ~308.04 K ops/sec | -1.81 %  |  ~89.55 K ops/sec | -2.40 %  |
+-------------------------------+-------------------+----------+-------------------+----------+
```
so even more pronounced! How do I compare against writing the binary format?

gapisback:
> This is what I see on the `perf-l3_write` branch:
> 
> ```
>     **** Performance comparison for NumClients=5, NumOps=5000000 (5 Million) ****
> +-------------------------------+-------------------+----------+-------------------+----------+
> | Run-Type                      | Server throughput | Srv:Drop | Client throughput | Cli:Drop |
> +-------------------------------+-------------------+----------+-------------------+----------+
> | Baseline - No logging         | ~313.73 K ops/sec |  0.00 %  |  ~91.75 K ops/sec |  0.00 %  |
> | L3-write() logging (no LOC)   | ~197.15 K ops/sec | -37.16 % |  ~49.13 K ops/sec | -46.46 % |
> | L3-logging default LOC        | ~314.68 K ops/sec |  0.30 %  |  ~92.17 K ops/sec |  0.45 %  |
> | L3-logging LOC-ELF            | ~308.04 K ops/sec | -1.81 %  |  ~89.55 K ops/sec | -2.40 %  |
> +-------------------------------+-------------------+----------+-------------------+----------+
> ```
> 
> so even more pronounced! How do I compare against writing the binary format?

This is very weird!

I had re-run this full set of perf tests w/o and w/ the bugfix for `write()` call. In [this previous comment](https://github.com/undoio/l3/pull/64#issuecomment-2138463324) of mine, I reported that I was seeing a drop of around 7.8%.

The default mode in this client-server program **is with the binary** `write()` call. See L391 in l3.c .

-----

I have not integrated support for `write_msg()` into the client-server program.

To exercise that do this: `$ make clean && CC=gcc LD=g++ make all-unit-tests`

This build step will produce this binary `build/release/bin/unit/l3-write-perf-test`

Run this with an arg specifying # of millions, and it will run thru both interfaces.

I showed some outputs earlier, in [this prev comment](https://github.com/undoio/l3/pull/64#issuecomment-2138450983) of mine, where the usage is described, too.

gregthelaw:
> This is very weird!

I'm not surprised TBH. Writing to the filesystem is (relatively) slow, even if it just hits the buffer cache. Prior to the bug fix we were writing 0 bytes, I wouldn't be surprised if the kernel optimises that away so it's minimal syscall entry/exit overhead. Now we're asking the OS to do a bunch of work.

It actually makes sense -- when you have a trivial amount of userspace code and some syscalls (as this test does) then really the only thing that counts is syscalls -- everything else is immaterial. The send/recv pair is 2 syscalls, and we add a 3rd syscall, and performance drops by about 1/3.

This is where I was coming from when I said the formatting of the message is unlikely to have a material impact once you've taken on the overheads of writing to the filesystem.

gregthelaw:
The `build/release/bin/unit/l3-write-perf-test` shows 
```
Initialized write() logging to '/tmp/l3-write-logging-test.dat'
1 Mil write() log msgs: 514 ns/msg (avg): /tmp/l3-write-logging-test.dat
Initialized write() logging to '/tmp/l3-write-logging-test.dat'
1 Mil write() log msgs: 505 ns/msg (avg): /tmp/l3-write-logging-test.dat
Initialized write() logging to '/tmp/l3-writemsg-logging-test.dat'
1 Mil write() log msgs: 592 ns/msg (avg): /tmp/l3-writemsg-logging-test.dat
```
So `write()` is about 15% faster than `writemsg()`, so I don't think it's worth it. i.e. I don't think it would ever make sense to have such a system requiring the post-processing for such minor gains.

I take the point that `writemsg` isn't really apples-to-apples with `l3_log()` but `write` compared to `l3_log()` just demonstrates that it's orders of magnitude slower writing a small binary blob to a file than it is to write it directly into some mmap'ed memory. Which isn't exactly breaking news, right? :)

gapisback:
> The `build/release/bin/unit/l3-write-perf-test` shows
> 
> ```
> Initialized write() logging to '/tmp/l3-write-logging-test.dat'
> 1 Mil write() log msgs: 514 ns/msg (avg): /tmp/l3-write-logging-test.dat
> Initialized write() logging to '/tmp/l3-write-logging-test.dat'

> ```
> 
Can you please re-run as follows:

`$ build/release/bin/unit/l3-write-perf-test 100`

This will run a warm-up run with 1M and will run perf-tests with both schemes with 100 Million msgs.

You can try with 200 also, but be warned -- the resulting log file runs into several GiBs.

gapisback:
> I take the point that `writemsg` isn't really apples-to-apples with `l3_log()` but `write` compared to `l3_log()` just demonstrates that it's orders of magnitude slower writing a small binary blob to a file than it is to write it directly into some mmap'ed memory. Which isn't exactly breaking news, right? :)


Seems like you are strongly leaning in favour of just doing `write()`-msg, which is fine by me.

How about this:

1. I will first get the bug-fix into /main -- which is non-controversial. Then nightly /main will start showing the correct drop in perf due to use of `write()` call.
2. In this branch, I will rework the client-server program to use this `write()`-msg interfaces.
3. You will need to re-run the full tests with baseline (i.e. nightly /main) and this new-fix version
4. Let's re-evaluate & record one last time the actual drop in perf of using binary write with formatted write.
5. Then, I will clean-up this change-set to cut-over completely to `write()`-msg form and we go with that.


You will have to do a few perf validation re-runs on your Linux workstation to get final numbers.
Does this work for you?

gregthelaw:
```
gel@onyx:~/src/l3$ build/release/bin/unit/l3-write-perf-test 100
Initialized write() logging to '/tmp/l3-write-logging-test.dat'
1 Mil write() log msgs: 524 ns/msg (avg): /tmp/l3-write-logging-test.dat
Initialized write() logging to '/tmp/l3-write-logging-test.dat'
100 Mil write() log msgs: 515 ns/msg (avg): /tmp/l3-write-logging-test.dat
Initialized write() logging to '/tmp/l3-writemsg-logging-test.dat'
100 Mil write() log msgs: 625 ns/msg (avg): /tmp/l3-writemsg-logging-test.dat
gel@onyx:~/src/l3$ 
gel@onyx:~/src/l3$ build/release/bin/unit/l3-write-perf-test 100
Initialized write() logging to '/tmp/l3-write-logging-test.dat'
1 Mil write() log msgs: 510 ns/msg (avg): /tmp/l3-write-logging-test.dat
Initialized write() logging to '/tmp/l3-write-logging-test.dat'
100 Mil write() log msgs: 504 ns/msg (avg): /tmp/l3-write-logging-test.dat
Initialized write() logging to '/tmp/l3-writemsg-logging-test.dat'
100 Mil write() log msgs: 602 ns/msg (avg): /tmp/l3-writemsg-logging-test.dat
gel@onyx:~/src/l3$ 
gel@onyx:~/src/l3$ build/release/bin/unit/l3-write-perf-test 100
Initialized write() logging to '/tmp/l3-write-logging-test.dat'
1 Mil write() log msgs: 523 ns/msg (avg): /tmp/l3-write-logging-test.dat
Initialized write() logging to '/tmp/l3-write-logging-test.dat'
100 Mil write() log msgs: 505 ns/msg (avg): /tmp/l3-write-logging-test.dat
Initialized write() logging to '/tmp/l3-writemsg-logging-test.dat'
100 Mil write() log msgs: 602 ns/msg (avg): /tmp/l3-writemsg-logging-test.dat
gel@onyx:~/src/l3$ 
```

gapisback:
> ```
> gel@onyx:~/src/l3$ build/release/bin/unit/l3-write-perf-test 100
> Initialized write() logging to '/tmp/l3-write-logging-test.dat'
> 1 Mil write() log msgs: 524 ns/msg (avg): /tmp/l3-write-logging-test.dat
> Initialized write() logging to '/tmp/l3-write-logging-test.dat'
> 100 Mil write() log msgs: 515 ns/msg (avg): /tmp/l3-write-logging-test.dat
> Initialized write() logging to '/tmp/l3-writemsg-logging-test.dat'
> 100 Mil write() log msgs: 625 ns/msg (avg): /tmp/l3-writemsg-logging-test.dat
> gel@onyx:~/src/l3$ 
> gel@onyx:~/src/l3$ build/release/bin/unit/l3-write-perf-test 100
> Initialized write() logging to '/tmp/l3-write-logging-test.dat'
> 1 Mil write() log msgs: 510 ns/msg (avg): /tmp/l3-write-logging-test.dat
> Initialized write() logging to '/tmp/l3-write-logging-test.dat'
> 100 Mil write() log msgs: 504 ns/msg (avg): /tmp/l3-write-logging-test.dat
> Initialized write() logging to '/tmp/l3-writemsg-logging-test.dat'
> 100 Mil write() log msgs: 602 ns/msg (avg): /tmp/l3-writemsg-logging-test.dat
> gel@onyx:~/src/l3$ 
> gel@onyx:~/src/l3$ build/release/bin/unit/l3-write-perf-test 100
> Initialized write() logging to '/tmp/l3-write-logging-test.dat'
> 1 Mil write() log msgs: 523 ns/msg (avg): /tmp/l3-write-logging-test.dat
> Initialized write() logging to '/tmp/l3-write-logging-test.dat'
> 100 Mil write() log msgs: 505 ns/msg (avg): /tmp/l3-write-logging-test.dat
> Initialized write() logging to '/tmp/l3-writemsg-logging-test.dat'
> 100 Mil write() log msgs: 602 ns/msg (avg): /tmp/l3-writemsg-logging-test.dat
> gel@onyx:~/src/l3$ 
> ```

For your runs on Linux-workstation: I don't see any variations in metrics between 1st and subsequent runs

So, taking the average of 3 runs, for 100 Mi msgs:

- `write()` takes 508 ns / msg
- `write()-msg` takes 610 ns / msg
- Avg of ~20% overhead in run-time perf due to formatting the message in-place before logging.

gapisback:
Hi, @gregthelaw - I have refactored and reorganized the 2 commits you were seeing before into the following 4 commits. These should lay down the stage well-enough for you to review and re-test the two schemes w/ the full-set of tests run by the `run-all-client-server-perf-tests` target:

1. [Commit-1](https://github.com/undoio/l3/pull/64/commits/a6e87ba8b384cf7227bd7b1356177cc7483a9257) [Perf] Support L3_LOGT_WRITE -> l3_log_write() for benchmarking.
2. [Commit-2](https://github.com/undoio/l3/pull/64/commits/af84c39e6eab33b1fafaba3d528f4db1d46c03b3): test.sh: Refactor into standalone write() logging test method
3. [Commit-3](https://github.com/undoio/l3/pull/64/commits/8258b7d3080fe200f1b44fa604a9395677537515): Add fprintf/write-logging perf unit-tests
4. [Commit-4](https://github.com/undoio/l3/pull/64/commits/ad36495b94c37a4a71891f37a8abda9aeaed5ae0): Integrate l3-write()-msg logging into client-server perf test

-----

I re-ran the whole batch of perf-tests on my Linux-VM, and got this output:

```
+ ./scripts/perf_report.py --file /tmp/test.sh.run-all-client-server-perf-tests.out
    **** Performance comparison for NumClients=5, NumOps=5000000 (5 Million) ****
+---------------------------------+-------------------+----------+-------------------+----------+
| Run-Type                        | Server throughput | Srv:Drop | Client throughput | Cli:Drop |
+---------------------------------+-------------------+----------+-------------------+----------+
| Baseline - No logging           |  ~95.42 K ops/sec |  0.00 %  |  ~21.77 K ops/sec |  0.00 %  |
| L3-logging (no LOC)             |  ~93.22 K ops/sec | -2.31 %  |  ~22.78 K ops/sec |  4.66 %  |
| L3-fast logging (no LOC)        |  ~93.88 K ops/sec | -1.62 %  |  ~21.35 K ops/sec | -1.91 %  |
| L3-fprintf() logging (no LOC)   |  ~90.74 K ops/sec | -4.90 %  |    21 K ops/sec   | -3.55 %  |
| L3-write() logging (no LOC)     |  ~82.90 K ops/sec | -13.12 % |  ~18.61 K ops/sec | -14.49 % |
| L3-write()-msg logging (no LOC) |  ~85.21 K ops/sec | -10.70 % |  ~18.69 K ops/sec | -14.14 % |
| L3-logging default LOC          |  ~99.35 K ops/sec |  4.11 %  |  ~23.48 K ops/sec |  7.87 %  |
| L3-logging LOC-ELF              | ~102.51 K ops/sec |  7.43 %  |  ~23.64 K ops/sec |  8.58 %  |
+---------------------------------+-------------------+----------+-------------------+----------+
+ set +x

test.sh: Fri May 31 10:13:33 AM PDT 2024 Completed all client(s)-server communication test [ 0h 7m 19s ].
```


Can you re-try `./test.sh run-all-client-server-perf-tests --clock-default $((1000 * 1000)) > ~/tmp/perf-write.v2.out 2>&1` and let's see what results you get on Linux box comparing both schemes.

----

**Default scheme choice**: 

You have been strongly in favour of `write()`-msg as the default, so I'm fine with that.

We can choose to retain, as an alternate option, the `write()`-logging interface, as well, which has a slightly better run-time gain, at the cost of having to do post-processing.

Let me know if you think this will just be clutter and more confusing, w/o any real user-experience gain. If so, I am ok with dismantling this 2nd approach, all its code-touch points and integrate the rest to /main.

Hopefully, with this new work, we can finalize on this perf-u-benchmarking and help me get this into /main.

src/l3.c
gregthelaw: Sure, it won't _write_ more than n-1 bytes, but it will _return_ the number of bytes _that it would have written if the buffer were big enough_.  From the man page:
```
       The  functions snprintf() and vsnprintf() do not write more than size bytes (including the terminating null byte ('\0')).  If the output was trun‐
       cated due to this limit, then the return value is the number of characters (excluding the terminating null byte) which would have been written  to
       the  final  string if enough space had been available.  Thus, a return value of size or more means that the output was truncated.
```
so if msg is, say, 266 bytes, we'll leak the next 10 bytes (probably garbage but who knows?) into the log.

gregthelaw (changes requested):
Almost there, but we need to resolve that comment on `snprintf`'s return value.

gregthelaw:
I was getting slightly odd results with 1,000,000 ops so went for 10,000,000:
```
+-------------------------------+-------------------+----------+-------------------+----------+
| Run-Type                      | Server throughput | Srv:Drop | Client throughput | Cli:Drop |
+-------------------------------+-------------------+----------+-------------------+----------+
| Baseline - No logging         | ~504.64 K ops/sec |  0.00 %  | ~207.60 K ops/sec |  0.00 %  |
| L3-logging (no LOC)           | ~499.65 K ops/sec | -0.99 %  | ~201.44 K ops/sec | -2.97 %  |
| L3-fast logging (no LOC)      | ~505.84 K ops/sec |  0.24 %  | ~207.25 K ops/sec | -0.17 %  |
| L3-fprintf() logging (no LOC) | ~478.51 K ops/sec | -5.18 %  | ~184.64 K ops/sec | -11.06 % |
| L3-write() logging (no LOC)   | ~346.93 K ops/sec | -31.25 % | ~106.35 K ops/sec | -48.77 % |
| L3-logging default LOC        | ~500.27 K ops/sec | -0.87 %  | ~204.79 K ops/sec | -1.35 %  |
| L3-logging LOC-ELF            | ~497.53 K ops/sec | -1.41 %  | ~200.12 K ops/sec | -3.60 %  |
+-------------------------------+-------------------+----------+-------------------+----------+
```

src/l3.c
```diff
@@ -416,7 +416,6 @@ l3_log_write_msg(const char *msgfmt, const uint64_t arg1, const uint64_t arg2)
 }
 
 /**
->>>>>>> e9e4dac (Fix)
```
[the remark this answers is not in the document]
gapisback: @gregthelaw - I reworked the commit deltas to remove this straggling chevron, which crept-in in the 1st commit in this sequence due to rebase v/s main and code shuffling.

src/l3.c
gapisback: Oh, I see ... I should have read the man page more carefully; sorry!

Thanks for pushing hard and close on this issue.

So, then, per this return value spec, your previous comment which suggested:

_I think you need max(msglen, sizeof msgbuf) as snprintf returns the number of bytes that would have been written -- i.e. can be more than `sizeof(msgbuf)._

... I think should say: need `min(msglen, sizeof(msgbuf))`, right?

sizeof(msgbuf) is hard-coded to 255.

If msglen is 100, we should write 100 bytes.
If msglen returned is 300, we should write 255 bytes.

I will make this change ... Ideally I need to whip-up a unit-test to verify these limits, but I'd like to do that post-check-in, to futz around with off-by-one limits testing using pytest. Hope you are ok with that approach.

src/l3.c
```diff
@@ -344,6 +374,49 @@ l3_log_mmap(const char *msg, const uint64_t arg1, const uint64_t arg2,
...
+    if (write(l3_log_fd, msgbuf, msglen) != msglen) {
+        fprintf(stderr, "%s(): write() failed. errno=%d\n",
+                __func__, errno);
+        return;
+    }
+    return;
+}
+
+/**
+>>>>>>> e9e4dac (Fix)
```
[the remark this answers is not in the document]
gapisback: Cleaned up and resolved in revised commit sequence under [upcoming] SHA f782329

gapisback:
@gregthelaw - I fixed the `snprintf()` return  value issue and the straggling chevrons.
Pushed a new set of rearranged commits to re-fire CI.

gapisback:
> I was getting slightly odd results with 1,000,000 ops so went for 10,000,000:
> 
> ```
> +-------------------------------+-------------------+----------+-------------------+----------+
> | Run-Type                      | Server throughput | Srv:Drop | Client throughput | Cli:Drop |
> +-------------------------------+-------------------+----------+-------------------+----------+
> | Baseline - No logging         | ~504.64 K ops/sec |  0.00 %  | ~207.60 K ops/sec |  0.00 %  |
> | L3-logging (no LOC)           | ~499.65 K ops/sec | -0.99 %  | ~201.44 K ops/sec | -2.97 %  |
> | L3-fast logging (no LOC)      | ~505.84 K ops/sec |  0.24 %  | ~207.25 K ops/sec | -0.17 %  |
> | L3-fprintf() logging (no LOC) | ~478.51 K ops/sec | -5.18 %  | ~184.64 K ops/sec | -11.06 % |
> | L3-write() logging (no LOC)   | ~346.93 K ops/sec | -31.25 % | ~106.35 K ops/sec | -48.77 % |
> | L3-logging default LOC        | ~500.27 K ops/sec | -0.87 %  | ~204.79 K ops/sec | -1.35 %  |
> | L3-logging LOC-ELF            | ~497.53 K ops/sec | -1.41 %  | ~200.12 K ops/sec | -3.60 %  |
> +-------------------------------+-------------------+----------+-------------------+----------+
> ```

Did you run with the latest set of commits?

I am not seeing the row for `write()-msg` logging interface. I would have expected to see this pair of rows in the output:

```
| L3-write() logging (no LOC)     |    987 ops/sec    | -0.10 %  |  ~17.98 K ops/sec | -7.83 %  |
| L3-write()-msg logging (no LOC) |    986 ops/sec    | -0.20 %  |  ~16.55 K ops/sec | -15.14 % |
```

gregthelaw:
**** Performance comparison for NumClients=5, NumOps=5000000 (5 Million) ****
+---------------------------------+-------------------+----------+-------------------+----------+
| Run-Type                        | Server throughput | Srv:Drop | Client throughput | Cli:Drop |
+---------------------------------+-------------------+----------+-------------------+----------+
| Baseline - No logging           | ~463.33 K ops/sec |  0.00 %  | ~173.28 K ops/sec |  0.00 %  |
| L3-logging (no LOC)             | ~455.55 K ops/sec | -1.68 %  | ~168.36 K ops/sec | -2.84 %  |
| L3-fast logging (no LOC)        | ~455.07 K ops/sec | -1.78 %  | ~168.03 K ops/sec | -3.03 %  |
| L3-fprintf() logging (no LOC)   | ~423.93 K ops/sec | -8.50 %  | ~147.81 K ops/sec | -14.70 % |
| L3-write() logging (no LOC)     | ~328.19 K ops/sec | -29.17 % |  ~97.79 K ops/sec | -43.57 % |
| L3-write()-msg logging (no LOC) | ~322.53 K ops/sec | -30.39 % |  ~95.33 K ops/sec | -44.98 % |
| L3-logging default LOC          | ~456.63 K ops/sec | -1.45 %  | ~169.57 K ops/sec | -2.14 %  |
| L3-logging LOC-ELF              | ~467.94 K ops/sec |  0.99 %  | ~176.66 K ops/sec |  1.95 %  |
+---------------------------------+-------------------+----------+-------------------+----------+

src/l3.c
gregthelaw: d'oh you're quite right - min not max obviously!

gregthelaw (approved):

PR #65, opened 2024-05-28
[L3] Add l3_log_deinit(): Close file opened.

gapisback:
Change server to deinit() L3's logging system, so contents are  reliably flushed to disk. Suppress call to syncfs() on Mac/OSX.

### DIFF
--- a/src/l3.c
+++ b/src/l3.c
@@ -216,6 +216,45 @@ l3_log_init(const l3_log_t logtype, const char *path)
 
 /**
  * ****************************************************************************
+ * l3_log_deinit() - De-initialize L3-logging sub-system, selecting the type
+ * of logging that was being performed. Finalize the file and close its handle.
+ * ****************************************************************************
+ */
+int
+l3_log_deinit(const l3_log_t logtype)
+{
+    int rv = 0;
+    switch (logtype) {
+      case L3_LOG_MMAP:         // L3_LOG_DEFAULT:
+        rv = munmap(l3_log, sizeof(*l3_log));
+        break;
+
+      case L3_LOG_FPRINTF:
+        fflush(l3_log_fh);
+        rv = fclose(l3_log_fh);
+        break;
+
+      case L3_LOG_WRITE:
+#if !defined(__APPLE__)
+        syncfs(l3_log_fd);
+#endif  // !__APPLE__
+        rv = close(l3_log_fd);
+        break;
+
+      default:
+        printf("Unsupported L3-logging type=%d\n", logtype);
+        return -1;
+    }
+    if (rv) {
+        fprintf(stderr, "Error closing L3 log-file for logging type '%s'"
+                        ", errno=%d\n",
+                l3_logtype_name(logtype), errno);
+    }
+    return rv;
+}
+
+/**
+ * ****************************************************************************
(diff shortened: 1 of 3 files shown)


src/l3.c
```diff
@@ -210,6 +210,45 @@ l3_log_init(const l3_log_t logtype, const char *path)
...
+{
+    int rv = 0;
+    switch (logtype) {
+      case L3_LOG_MMAP:         // L3_LOG_DEFAULT:
+        rv = munmap(l3_log, sizeof(*l3_log));
+        break;
+
+      case L3_LOG_FPRINTF:
+        fflush(l3_log_fh);
+        rv = fclose(l3_log_fh);
```
gapisback, line 230: I manually verified that this sequence works fine ...

src/l3.c
```diff
@@ -210,6 +210,45 @@ l3_log_init(const l3_log_t logtype, const char *path)
...
+        break;
+
+      case L3_LOG_WRITE:
+#if !defined(__APPLE__)
+        syncfs(l3_log_fd);
+#endif  // !__APPLE__
+        rv = close(l3_log_fd);
```
gapisback, line 237: But somehow ..., `syncfs()` followed by `close()` seems to not flush the contents to disk.

I am still getting files that are 0 bytes.

We should investigate. Can you do a manual test on your box ... and see if this is [not] working as I described?

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -408,6 +408,12 @@ main(int argc, char *argv[])
     if (msgctl(serverId, IPC_RMID, NULL) == -1) {
         errExit("msgctl");
     }
+
+    // Close L3-logging upon exit.
+#if L3_ENABLED
+    l3_log_deinit(logtype);
```
gapisback, line 414: A nice single-exit point `deinit()` hook.

src/l3.c
gregthelaw: I had a quick look. At least for `single-file-C-program` it is not calling `l3_log_init()` but `l3_init()` so this means that `l3_log_fd` is `-1`.

I had a quick look with Undo to see if I can diagnose it. Worked a treat :)  The problem is in `l3_log_write()`
```
      if (write(l3_log_fd, lep, sizeof(*lep) != sizeof(*lep))) {
```
the first closing paren is in the wrong place, so it's passing 0 as the length argument to `write()`. i.e. the fix is simple:
```
diff --git a/src/l3.c b/src/l3.c
index 133fef6..e3d87c1 100644
--- a/src/l3.c
+++ b/src/l3.c
@@ -426,7 +426,7 @@ l3_log_write(const char *msg, const uint64_t arg1, const uint64_t arg2)
     lep->arg1 = arg1;
     lep->arg2 = arg2;
 
-    if (write(l3_log_fd, lep, sizeof(*lep) != sizeof(*lep))) {
+    if (write(l3_log_fd, lep, sizeof(*lep)) != sizeof(*lep)) {
         fprintf(stderr, "%s(): write() failed. errno=%d\n",
                 __func__, errno);
         return;
```
This sort of error is just too easy in C!

gregthelaw (changes requested):

PR #66, opened 2024-05-28
[Spdlog] Add spdlog sample program, Makefile and CI build.yml rules

gapisback:
This commit introduces a very basic spdlog-tutorial program.

- Add use-cases/spdlog-Cpp-program/test-main.cpp
- Makefile:
   - Add build-rules to build this source on Linux, Mac/OSX and Docker
   - New target spdlog-cpp-program; needs CC=g++ LD=g++
   - Handle env-var RUN_ON_CI, to fix SPD_INCLUDE_ROOT path that works on CI-Mac/OSX machines, as installed by `brew install`

- test.sh:
   - Add test-build-and-run-spdlog-sample

- CI/build.yml:
   - Install required pre-requisite packages, spdlog and fmt
   - Customize install env for Linux & Mac/OSX
   - Provide special-invocation env-var: RUN_ON_CI=1

------

## [Perf] Extend client-server perf tests to run with spdlog-logging

This commits injects spdlog-logging into the existing client-server performance u-benchmarking programs.

Usage:

```
make clean && CC=g++ LD=g++ L3_ENABLED=0 L3_LOGT_SPDLOG=1 make client-server-perf-test  # C++ spdlog logging
make clean && CC=g++ LD=g++ L3_ENABLED=0 L3_LOGT_SPDLOG=2 make client-server-perf-test  # C++ spdlog backtrace
```

- Makefile:
      - Extend build-rules to compile programs with above env-vars
      - Cleanup CPPFLAGS definitions to get all programs' builds working.
      - Deal with .c files that are compiled with g++ for spdlog usage.

- test.sh:
      - Add test-build-and-run-client-server-perf-test-spdlog
      - Add above test-method to run-all-client-server-perf-tests
      - CI: New tests are automatically executed as CI runs prev `run-all` test-method.

     - Fix Mac/OSX porting issues for spdlog related perf-tests.

 - svmsg_file_server.c: Add conditional code to call spdlog interfaces to initiate
      logging. Invoke logging to a file and also backtrace logging.

      Small set of changes to compile .c file using C++ compiler.

### DIFF
--- a/use-cases/client-server-msgs-perf/svmsg_file_server.c
+++ b/use-cases/client-server-msgs-perf/svmsg_file_server.c
@@ -69,6 +69,18 @@ Usage: ./svmsg_file_server --help
 #include <time.h>
 #include <assert.h>
 #include <getopt.h>     // For getopt_long()
+
+#ifdef __cplusplus
+#  if defined(L3_LOGT_SPDLOG) or defined(L3_LOGT_SPDLOG_BACKTRACE)
+
+#   include <iostream>
+#   include "spdlog/spdlog.h"
+#   include "spdlog/sinks/basic_file_sink.h"
+    using namespace std;
+
+#  endif    // L3_LOGT_SPDLOG
+#endif  // __cplusplus
+
 #include "svmsg_file.h"
 #include "l3.h"
 #include "size_str.h"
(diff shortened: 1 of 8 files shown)


use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -285,7 +334,7 @@ main(int argc, char *argv[])
             break;                      /* ... so terminate loop */
         }
 
-        switch (req.mtype) {
+        switch ((int) req.mtype) {
```
gregthelaw, line 337: How come this is necessary?

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -323,17 +372,28 @@ main(int argc, char *argv[])
...
+                             "Increment: ClientID={}, Counter{}.",
+                             resp.clientId, resp.counter);
+
+#  elif defined(L3_LOGT_SPDLOG_BACKTRACE)
+
+            spdlog::debug("Server spdlog-Backtrace: "
+                         "Increment: ClientID={}, Counter={}.",
+                         resp.clientId, resp.counter);
+
+#elif L3_ENABLED
```
gregthelaw, line 387: I note an inconsistently here -- for speedlog the macro need only be defined, but for L3 etc it must be defined to non-zero. (Not a big deal.)

gregthelaw (approved):
Minor comments only

use-cases/client-server-msgs-perf/svmsg_file_server.c
gapisback: This is a `.c` file which now needs to be compiled with C++ compiler as we are injecting spdlog calls into it. Spdlog is a C++ library.

I think clang on Mac/OSX complains if this `int` typecasting is not there. I forgot the exact compiler error.

use-cases/client-server-msgs-perf/svmsg_file_server.c
gapisback: I understand ... here are my reasons.

Actually, in Makefile all these #if'defs are issued with `-D<flag> ` clause, so, technically, I could use `#if defined(L3_ENABLED)` also.

For `L3_ENABLED`, we have some invocations like:

```
CC=gcc LD=g++ L3_ENABLED=0 make client-server-perf-test  # Baseline

L3_ENABLED=0 L3_LOGT_SPDLOG=1 make client-server-perf-test  # C++ spdlog logging
```

Where one has to explicitly turn OFF `L3_ENABLED` by setting it to 0.

So, there the use of `#if L3_ENABLED` would not be fully indicative of the fact that sometimes this env-var setting at `Makefile`-context is acting like a boolean.

Env-var setting `L3_LOGT_SPDLOG=1,2` translates to one of `-DL3_LOGT_SPDLOG` or `-DL3_LOGT_SPDLOG_BACKTRACE` . Here, `#if defined()` seemed appropriate.

I am not an expert in figuring out these conventions & rules for when to use which form.

I'd like to get this feature-support in and we can revisit holistically the right code-patterns, starting from env-vars thru Makefile `-D<flags>` and fix things as needed.

PR #71, opened 2024-06-07
[Perf] Refactor svmsg_file_server.c to prepare for threaded-server execution.

gapisback:
This commit lays down the ground work to run the server program of the client-server performance u-benchmarking toolkit as multi-threaded execution. The goal is to invoke and measure the overhead of concurrent logging performed by multiple server threads in response to messages received from n-concurrent clients.

- Add support for `--num-server-threads <n>` argument, but currently not invoked. Default is num_threads = 1

- Break-out inline code into `svr_op_*()` work-horse functions to implement the important msg-types. Driver fn is `svr_proc_process_msg()`, which is currently invoked by a single server-thread.

These changes lay-down enough pthread-create/join code to make it easier to run, stabilize and re-benchmark overheads of logging using a multi-threaded server.


Makefile
```diff
@@ -730,6 +730,8 @@ ifeq "$(BUILD_VERBOSE)" "1"
 endif
 
 $(BINDIR)/$(USE_CASES)/svmsg_file_client: $(CLIENT_SERVER_CLIENT_MAIN_OBJ) $(CLIENT_SERVER_NON_MAIN_OBJS)
+
+$(BINDIR)/$(USE_CASES)/svmsg_file_server: LIBS += -lpthread
```
gapisback, line 734: Need this to use pthread-APIs.

use-cases/client-server-msgs-perf/svmsg_file.h
```diff
@@ -38,7 +38,8 @@
 /**
  * Message Types for request / response messages sent from server to client
  */
-typedef enum {
+typedef long req_resp_type_t;
```
gapisback, line 41: Cleaning up an existing inconsistency.

As per the [msgsnd() man](https://www.man7.org/linux/man-pages/man3/msgsnd.3p.html) page, we need the message structure to be defined as:

```
       The application shall ensure that the argument msgp points to a
       user-defined buffer that contains first a field of type long
       specifying the type of the message, and then a data portion that
       holds the data bytes of the message. The structure below is an
       example of what this user-defined buffer might look like:

           struct mymsg {
               long   mtype;       /* Message type. */
               char   mtext[1];    /* Message text. */
           }
```

I have no idea how things used to work before, but this seems to be the right thing to do.

Kindly do cross-check and verify that this change is kosher.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -106,6 +107,13 @@ typedef struct client_info {
 
 Client_info ActiveClients[MAX_CLIENTS];
 
+/*
+ * Configuration for each server thread.
+ */
+typedef struct {
+    int     server_id;  // Server's msgq-ID.
+} svr_thread_config;
```
gapisback, line 115: Arg-struct to invoke `pthread_create()`

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -123,6 +131,16 @@ Client_info ActiveClients[MAX_CLIENTS];
...
 static unsigned int   NumActiveClients    = 0;
 
+/**
+ * On MacOSX, only the CLOCK_THREAD_CPUTIME_ID clock has a resolution of 1ns.
+ * On Linux, all clocks seem to have 1 ns resolution, so use the CLOCK_REALTIME
+ * clock as the default clock to use. This way, the throughput metric computed
+ * as (nops/sec) is consistent with normal expectation of 'sec' as real time
+ * seconds.
+ */
+int Clock_id = CLOCK_REALTIME;
```
gapisback, line 141: This variable used to be in `main()` but needs to be referenced by child threads.

So made it global and renamed to Upper-case, to keep a convention I'm maintaining of naming global variables with 1st char upper-case.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -136,18 +154,33 @@ struct option Long_options[] = {
     , { "clock-thread-cputime-id"   , no_argument         , NULL, 't'}
 
     // Options that need arguments
+    , { "num-server-threads"        , required_argument   , NULL, 'n'}
```
gapisback, line 157: Introduced for now, but is not being invoked in any test execution.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -136,18 +154,33 @@ struct option Long_options[] = {
...
 };
 
-const char * Options_str = "o:dhmprt";
+const char * Options_str = "no:dhmprt";
 
 // Useful macros
 #define ARRAY_LEN(arr)  (sizeof(arr) / sizeof(*arr))
 
+#if defined(L3_LOGT_SPDLOG)
+std::shared_ptr<spdlog::logger> Spd_logger; // = spdlog::basic_logger_mt();
```
gapisback, line 168: Made it global, so that the code in the work-horse function implementing the operation, i.e., `svr_op_incr()` can still reference this spdlog-logging handle when running tests in that build-mode.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -296,33 +319,147 @@ main(int argc, char *argv[])
...
+    // Invoke pthread_create() to create n-threads ...
+    svr_thread_config   svr_config = {
+                            .server_id = serverId
+                        };
+    pthread_t   thread_ids[num_threads];
 
     struct timespec ts0 = {0};
-    struct timespec ts1 = {0};
 
+    if (svr_start_threads(thread_ids, num_threads, (void *) &svr_config)) {
```
gapisback, line 344: Currently, this will start just 1 server-thread, essentially the same code as before, but serialized through one thread.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -296,33 +319,147 @@ main(int argc, char *argv[])
...
+        errExit("Server thread creation failed.");
+    }
 #if !defined(__APPLE__)
-    if (clock_gettime(clock_id, &ts0)) {    // ***** Timing begins
+    if (clock_gettime(Clock_id, &ts0)) {        // ***** Timing begins
```
gapisback, line 348: **NOTE**: The clock-begin is happening -after- the server thread has been started.

In future, when this is enhanced to actually support n-threads, we will start-up the threads, each of which will sit around waiting in `msgrecv()`. The timer-begin will happen sometime after that.

So, there is a bit of fuzziness in exactly when the clock will begin, but I think it should average out over millions of requests processed.

Let me know if you think this timing-begin logic needs to be tightened more.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -296,33 +319,147 @@ main(int argc, char *argv[])
...
+#else   // __cplusplus
+
+#if !defined(L3_ENABLED)
+    svr_clock_calibrate();
+#endif  // L3_ENABLED
+
+#endif  // __cplusplus
+
+    exit(EXIT_SUCCESS);
+}
```
gapisback, line 416: No change from L362 thru here.

Just a diff-block that shows up due to code modularization and refactoring.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -296,33 +319,147 @@ main(int argc, char *argv[])
...
+    return rv;
+}
+
+/**
+ * -----------------------------------------------------------------------------
+ * svr_proc_process_msg() - for()-ever loop to process messages from clients.
+ * -----------------------------------------------------------------------------
+ */
+void *
+svr_proc_process_msg(void *cfg)
```
gapisback, line 453: Chunk of for()-ever looping code modularized out from `main()` into this worker function.

Each `case:` block is further modularized out into stand-alone functions. (That may have not been necessary, but I had done that in earlier dev rounds, while trying to create stubs for worker threads to execute.)

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -337,73 +474,21 @@ main(int argc, char *argv[])
...
-                        ". (L3-fast-log)",
-                        resp.clientId, resp.counter);
-#  else
-            l3_log("Server msg: Increment: ClientID=%d, Counter=%" PRIu64 ".\n",
-                   resp.clientId, resp.counter);
-#  endif    // L3_FASTLOG_ENABLED
-
-#endif // L3_ENABLED
-
+            if (svr_op_incr(&req)) {
```
gapisback, line 483: Logging and `msgsnd()` back to the client is pushed down into this routine.

gregthelaw (approved):

PR #75, opened 2024-06-10
[Perf] Support multi-threaded server execution. Update test.sh perf methods.

gapisback:
This commit supports use of `--num-server-threads n` arg (to the `svmsg_file_server program`) with n > 1 threads. This will invoke n-server-side threads, to receive and process the messages received from the client program.

The objective is to measure any perf-impact due to concurrent logging of messages performed by multiple server threads.

- svmsg_file_server.c:
  - Prev commit already support notion of num_threads. With this change, we are now able to execute tests using more than 1 server-thread.

  - Use atomic int support to manage updates to # client-connections globals. Adjust code to work with both C & C++ compilation.

    Fortunately, due to the simplistic client-specific way messages are received & processed, we needed no new synchronization methods between server threads. Atomic incr / decrement of book-keeping variables was sufficient.

- Each server-thread reports the # of messages it processed. Update scripts/perf_report.py to report avg # of messages processed by each thread, to check that all messages are being uniformly distributed across all threads.

- Add scripts/ipcrm.sh -- script to clean-up left-over msg queues. In some runs on AWS instances, we were seeing server seg-faults in perf-runs with spdlog logging. This causes the entire test execution to hang. test.sh is updated locally to invoke this script to clean-up left-over message queues. (The actual source of the seg-fault is still tbd.)

------

NOTE TO THE REVIEWER:

@gregthelaw -- You may want to test the new feature of server-threads out on your Linux workstation as follows:

- Single-threaded execution:

```
./test.sh run-all-client-server-perf-tests --clock-default $((1000 * 1000)) > /tmp/perf.1thread.out 2>&1
```

- Multi-threaded execution:

```
./test.sh run-all-client-server-perf-tests --clock-default $((1000 * 1000)) 2 > /tmp/perf.2threads.out 2>&1
```

If you can run these and archive the summary report in a comment, it will be useful record to archive.

### DIFF
--- a/scripts/ipcrm.sh
+++ b/scripts/ipcrm.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+# ##############################################################################
+# Release unused shared segments.
+#
+# History:
+#	May-2024	Written; developing perf u-benchmarking scripts for L3-logging
+# ##############################################################################
+
+set -e -u -o pipefail
+
+Me=$(basename "$0")
+
+for shmid in $(ipcs -m | grep -w '0' | awk '{print $2}'); do
+
+	echo "$Me: Releasing shmid $shmid"
+	ipcrm -m ${shmid}
+done
+
+for msqid in $(ipcs -q | grep -v -i -w -E '1|Message|msqid' | awk '{print $2}'); do
+    echo "${msqid}"
+	ipcrm -q ${msqid}
+done
+
+ipcs -am
--- a/scripts/perf_report.py
+++ b/scripts/perf_report.py
@@ -10,6 +10,7 @@ Copyright (c) 2024
 import sys
 import argparse
 from collections import OrderedDict
+from collections import namedtuple
 from prettytable import PrettyTable
 
 ###############################################################################
@@ -28,7 +29,7 @@ def do_main(args:list):
     Main method that drives parsing the input file to generate the report.
     Each line is of the format:
      - Multiple comma-separated fields
-     - Each field may is '='-separated, like so; "<metric>=<value> (<text>)"
+     - Each field is '='-separated, like so; "<metric>=<value> (<text>)"
 
 # pylint: disable=line-too-long
 
@@ -43,10 +44,33 @@ L3-logging (no LOC), NumClients=5, NumOps=5000000 (5 Million), Server throughput
     perf_file = parsed_args.file
     lctr = 0
 
+    # Key: Server-thread-count 'NumThreads=1'; Value: metrics_by_run{}
+    metrics_by_threads = OrderedDict()
+
+    # Key: Server-thread-count 'NumThreads=1'; Value: namedtuple (svr_value, cli_value)
+    baseline_tuple = namedtuple('baseline_tuple', 'svr_value cli_value')
+    baseline_by_threads = OrderedDict()
+
+    # Key: Run-type 'Baseline - No logging'; Value: [ <metrics> ]
     metrics_by_run = OrderedDict()
     heading_row = []
-    base_svr_value = None   # baseline server-throughput
-    base_cli_value = None   # baseline client-throughput
+
+    # Setup some defaults to avoid pylint errors.
+    nclients_field = 'NumClients=5'
+    nops_field = 'NumOps=5000'
+
+    # We expect a consistent format of one-line summary messages arranged by:
+    #
+    #   - Run-Type-1: < Set of metrics for NumThreads=1 >
+    #   - Run-Type-1: < Set of metrics for NumThreads=2 >
+    #   - Run-Type-1: < Set of metrics for NumThreads=4 >
+    #   - Run-Type-1: < Set of metrics for NumThreads=8 >
+    #
+    #   - Run-Type-2: < Set of metrics for NumThreads=1 >
+    #   - Run-Type-2: < Set of metrics for NumThreads=2 >
+    #   - Run-Type-2: < Set of metrics for NumThreads=4 >
+    #   - Run-Type-2: < Set of metrics for NumThreads=8 >
+    # and so on ...
 
     with open(perf_file, 'rt', encoding="utf-8") as file:
         for line in file:
@@ -56,26 +80,46 @@ L3-logging (no LOC), NumClients=5, NumOps=5000000 (5 Million), Server throughput
                 (nclients_field, nops_field, nthreads_field, svr_metric, cli_metric, thread_metric) = parse_perf_line_names(line)
                 heading_row = [ 'Run-Type', svr_metric, cli_metric, thread_metric ]
 
-            (run_type, svr_value, svr_str, cli_value, cli_str, thread_str) = \
+            (run_type, nthreads_field, svr_value, svr_str, cli_value, cli_str, thread_str) = \
                 parse_perf_line_values(line)
 
+            # Create hash-entry for baseline server/client metrics for this thread-count
             if run_type.startswith('Baseline'):
-                base_svr_value = svr_value
-                base_cli_value = cli_value
+                # svr_value - baseline server-throughput
+                # cli_value - baseline client-throughput
+                baseline_by_threads[nthreads_field] = baseline_tuple(svr_value, cli_value)
 
-            metrics_by_run[run_type] = [svr_value, cli_value, svr_str, cli_str, thread_str ]
+            # DEBUG: print(f"{run_type}, {thread_str}")
 
+            # ------------------------------------------------------------------
+            # We want to distribute the metrics for each run-type across the
+            # map of metrics arranged by #-of-server-threads.
+            # - If this thread-count is a new one, add the dictionary-entry
+            #   { run_type: [ <metrics> ] } to the metrics tracked per thread.
+            # - Otherwise, update the metrics previously seen for previous
+            #   run-types for this thread-count with the metrics of the
+            #   new run-type seen.
+            #
+            # This way, we will be rearranging the one-line metrics rows
+            # into a dictionary of dictionaries, arranged by thread-count.
+            # ------------------------------------------------------------------
+            by_nthreads_dict = OrderedDict()
+            if nthreads_field not in metrics_by_threads:
+                metrics_by_threads[nthreads_field] = { run_type: [svr_value, cli_value, svr_str, cli_str, thread_str ] }
+            else:
+                by_nthreads_dict = metrics_by_threads[nthreads_field]
+                by_nthreads_dict.update( {run_type: [svr_value, cli_value, svr_str, cli_str, thread_str ] } )
+                metrics_by_threads[nthreads_field] = by_nthreads_dict
+
+            pr_metrics_by_thread(metrics_by_threads)
             lctr += 1
 
-    # DEBUG: pr_metrics(metrics_by_run)
 
-    if base_svr_value is None or base_cli_value is None:
-        print("Error: Throughputs metrics for baseline run are not known.")
-        sys.exit(1)
+    # DEBUG: pr_metrics(metrics_by_run)
+    # DEBUG: pr_metrics_by_thread(metrics_by_threads)
 
     # pylint: disable-next=line-too-long
-    print(f"    **** Performance comparison for {nclients_field}, {nops_field}, {nthreads_field} ****")
-    gen_perf_report(heading_row, metrics_by_run, base_svr_value, base_cli_value)
+    gen_perf_report_by_threads(heading_row, metrics_by_threads, baseline_by_threads, nclients_field, nops_field)
 
 # #############################################################################
 def parse_perf_line_names(line:str) -> tuple:
@@ -121,11 +165,13 @@ def parse_perf_line_values(line:str) -> tuple:
         svr_str     : str   : '(~91.79 K) ops/sec'
         cli_value   : int   : 20938
         cli_str     : str   : '(~20.93 K) ops/sec'
-        thread_str  : str   : '(25 K) ops/thread'
+        thread_str  : str   : '25 K'
     """
     # pylint: disable-next=unused-variable,line-too-long
-    (run_type, ct_unused, nops_unused, nthreads_unused, svr_field, cli_field, time_field_unused, nops_per_thread) = line.split(',')
+    (run_type, ct_unused, nops_unused, nthreads_field, svr_field, cli_field, time_field_unused, nops_per_thread) = line.split(',')
+
     run_type = run_type.lstrip().rstrip()
+    nthreads_field = nthreads_field.lstrip().rstrip()
 
     (svr_value_str, svr_str) = split_field_value_str(svr_field)
     (cli_value_str, cli_str) = split_field_value_str(cli_field)
@@ -136,7 +182,8 @@ def parse_perf_line_values(line:str) -> tuple:
     svr_value = int(svr_value_str)
     cli_value = int(cli_value_str)
 
-    return (run_type, svr_value, svr_str, cli_value, cli_str, thread_str)
+    print(f"{run_type}, {nthreads_field}")
+    return (run_type, nthreads_field, svr_value, svr_str, cli_value, cli_str, thread_str)
 
 
 # #############################################################################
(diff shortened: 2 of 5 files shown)


scripts/ipcrm.sh
```diff
@@ -0,0 +1,24 @@
...
+
+for shmid in $(ipcs -m | grep -w '0' | awk '{print $2}'); do
+
+	echo "$Me: Releasing shmid $shmid"
+	ipcrm -m ${shmid}
+done
+
+for msqid in $(ipcs -q | grep -v -i -w -E '1|Message|msqid' | awk '{print $2}'); do
+    echo "${msqid}"
+	ipcrm -q ${msqid}
```
gapisback, line 21: Script to cleanup leftover message queues in case server runs into seg-fault due to some bug.

Have seen this occur at least once in test-runs, so productizing this script to make `test.sh` execution a bit more reliable.

scripts/perf_report.py
```diff
@@ -52,17 +52,18 @@ def do_main(args:list):
         for line in file:
             if lctr == 0:
 
-                (nclients_field, nops_field, svr_metric, cli_metric) = parse_perf_line_names(line)
-                heading_row = [ 'Run-Type', svr_metric, cli_metric ]
+                # pylint: disable-next=line-too-long
+                (nclients_field, nops_field, nthreads_field, svr_metric, cli_metric, thread_metric) = parse_perf_line_names(line)
+                heading_row = [ 'Run-Type', svr_metric, cli_metric, thread_metric ]
```
gapisback, line 57: Changes to report # of server-threads in effect.

scripts/perf_report.py
```diff
@@ -133,11 +145,11 @@ def split_field_value_str(metric_field:str) -> (int, str):
     Specialized string-'split' method.
     Common case:
       Input:  'Server throughput=53856 (~53.85 K) ops/sec'
-      Returns:    (53856, ~53.85 K ops/sec')
+      Returns:    '~53.85 K ops/sec'
 
     Lapsed case:
       Input:  'Server throughput=989 () ops/sec'
-      Returns:    (53856, 989 ops/sec')
+      Returns:    '989 ops/sec'
```
gapisback, line 152: Here and on L148 above, cleanup the field as returned. Bug in old comments.

tests/pytests/perf_report_test.py
```diff
@@ -26,11 +26,11 @@
 # Canonical string generated from perf u-benchmarking scripts, used
 # to test out parsing methods.
 # pylint: disable=line-too-long
-MSG1 = 'Baseline - No logging, NumClients=5, NumOps=5000000 (5 Million), Server throughput=91795 (~91.79 K) ops/sec, Client throughput=20938 (~20.93 K) ops/sec, elapsed_ns=54469084570 (~54.46 Billion) ns'
+MSG1 = 'Baseline - No logging, NumClients=5, NumOps=5000000 (5 Million), NumThreads=1, Server throughput=91795 (~91.79 K) ops/sec, Client throughput=20938 (~20.93 K) ops/sec, elapsed_ns=54469084570 (~54.46 Billion) ns, NumOps/thread=5000 (5 K)'
```
gapisback, line 29: Update test msg-line format to conform to what's being spit out by the server in summary-line format mode.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -107,7 +108,13 @@ Client_info ActiveClients[MAX_CLIENTS];
...
  */
 typedef struct {
-    int     server_id;  // Server's msgq-ID.
+    int         server_id;  // Server's msgq-ID.
+
+    // On Mac/OSX, pthread_self() returns a struct, not an ID. So, to get
+    // code compiling, print the thread's index in the threads[] array.
+    int         thread_idx;
+    uint64_t    num_ops;    // Processed by this thread
+
```
gapisback, line 117: Thread config struct changed so we can compile code on Mac.
Also on L 116, report back the # of messages processed by each thread.

In summary report, this field is reported as avg. # of messages / thread -- which shows us quickly whether messages are being distributed reasonably uniformly across server threads.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -456,8 +494,19 @@ svr_proc_process_msg(void *cfg)
...
+    uint64_t tid = svr_config->thread_idx;
+#else
+    pthread_t tid = pthread_self();
+#endif  // __APPLE__
+
+    uint64_t    num_ops = 0;
     for (;;) {
 
+        if (NumActiveClientsHWM && (NumActiveClients == 0)) {
+            goto end_forever_loop;
```
gapisback, line 507: Each thread will 'block' in `msgrcv()` on L510, below.

So, this check here is an early-check to see if all clients have drained out. If so, this thread can also exit.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -489,17 +539,38 @@ svr_proc_process_msg(void *cfg)
...
 
-            if (NumActiveClients == 0) {
+                if (NumActiveClients == 0) {
+                    req.mtype = REQ_MT_EXIT;
+
+                    // Inform last remaining server-thread that this is
+                    // a final-exit from a server-thread, and not from a real
+                    // client.
+                    req.client_idx = -1;
+                    if (msgsnd(serverId, &req, REQ_MSG_SIZE, 0) == -1) {
```
gapisback, line 570: This is a slightly hacky way of handling the case when last-client drains out of the system.

When that client sends the `REQ_MT_EXIT` message, the thread handling it will detect that `NumActiveClients` has gone down to 0.

It's likely that one of the other server threads is waiting for a message.

The code here "informs" the other thread to also exit. The receiving logic is above on L143.

---
This scheme works when tested with 2 or 4 threads and 5 clients.

It's not a general purpose fix, yet, but something to get tests working for now.
Once I get access to an AWS instance with more cores, will be able to re-test this exit condition handling with larger number(s) of clients & threads. And if necessary, will tighten this area to be robust.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -543,22 +615,26 @@ svr_op_ct_init(requestMsg *req)
     resp.clientId = req->clientId;   // For this client
 
     // Use the next free slot w/o trying to recycle existing free slots.
-    resp.client_idx = NumActiveClientsHWM;
+    resp.client_idx = NumActiveClientsHWM++;
```
gapisback, line 618: Atomically increment `NumActiveClientsHWM` and then use old-value of `NumActiveClientsHWM` as this client's index .

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -543,22 +615,26 @@ svr_op_ct_init(requestMsg *req)
...
     memset(clientp, 0, sizeof(*clientp));
 
     clientp->clientId   = req->clientId;
-    clientp->client_idx = NumActiveClientsHWM;
+    clientp->client_idx = resp.client_idx;
     clientp->client_ctr = req->counter;
     clientp->last_mtype = req->mtype;
 
     NumActiveClients++;
-    NumActiveClientsHWM++;
```
gapisback, line 559: NOTE: There is a potential race on prev L630 where `NumActiveClientsHWM` is incremented earlier, but `NumActiveClients` is incremented above.

In theory, both should be incremented in lock-step.

In practice, it should not matter critically if the client's index and `NumActiveClients` don't quite sync-up.

I have not found anything untoward so far with this approach, so will let this be, to see how things fare in further testing.

use-cases/client-server-msgs-perf/svmsg_file_server.c
```diff
@@ -124,8 +131,13 @@ typedef struct {
...
  *   the slot(s) freed by clients who exited previously.
  */
-static unsigned int   NumActiveClientsHWM = 0;
-static unsigned int   NumActiveClients    = 0;
+#ifdef __cplusplus
+static std::atomic<int> NumActiveClientsHWM{0};
+static std::atomic<int> NumActiveClients{0};
+#else   // __cplusplus
+static _Atomic int NumActiveClientsHWM = 0;
+static _Atomic int NumActiveClients    = 0;
```
gapisback, line 139: The main change needed to synchronize between threads is the conversion of these counter variables to atomic.

The rest seemed to fall in-place fairly (& surprisingly) easily.

Please do cross-check the logic changes around these fields, to make sure I am not making any basic error which will introduce race conditions.

scripts/perf_report.py
```diff
@@ -56,26 +80,46 @@ def do_main(args:list):
...
+            # This way, we will be rearranging the one-line metrics rows
+            # into a dictionary of dictionaries, arranged by thread-count.
+            # ------------------------------------------------------------------
+            by_nthreads_dict = OrderedDict()
+            if nthreads_field not in metrics_by_threads:
+                metrics_by_threads[nthreads_field] = { run_type: [svr_value, cli_value, svr_str, cli_str, thread_str ] }
+            else:
+                by_nthreads_dict = metrics_by_threads[nthreads_field]
+                by_nthreads_dict.update( {run_type: [svr_value, cli_value, svr_str, cli_str, thread_str ] } )
+                metrics_by_threads[nthreads_field] = by_nthreads_dict
```
gapisback, line 112: Major rework is in this block, to rearrange the metrics generated for each run-type to distribute them across a dictionary keyed on "# of server threads."

See the sample outputs posted in discussion item #79 .

scripts/perf_report.py
```diff
@@ -173,6 +220,33 @@ def split_field_value_str(metric_field:str) -> (int, str):
 
     return (svr_value, svr_str)
 
+# #############################################################################
+def gen_perf_report_by_threads(heading:list, metrics_by_threads:dict, baseline_by_threads:dict,
```
gapisback, line 224: New driver routine, which calls existing report generation routine, on L247, below.

use-cases/client-server-msgs-perf/svmsg_file_server.c
gregthelaw: Looks good!

use-cases/client-server-msgs-perf/svmsg_file_server.c
gregthelaw: Is this a time-to-check-to-time-of-use race? Specifically, what happens if `NumActiveClients` becomes 0 after the check?

use-cases/client-server-msgs-perf/svmsg_file_server.c
gregthelaw: Would it be simpler just to `exit()` in this case? i.e. don't bother trying to wake the other threads, just nuke the whole the thing from orbit.

use-cases/client-server-msgs-perf/svmsg_file_server.c
gregthelaw: I think this is wrong for C code. I don't know C++ but assume the ++ operator will do the right thing, but for C code you want something like `__sync_fetch_and_add()`. To keep things simpler you might want to just use that throughout and not bother the C++ atomics stuff. (I don't feel strongly either way on that though, just a thought.)

use-cases/client-server-msgs-perf/svmsg_file_server.c
gregthelaw: Agree, I don't see a problem with `NumActiveClientsHWM` and `NumActiveClients` being out of sync.

gregthelaw (changes requested):

use-cases/client-server-msgs-perf/svmsg_file_server.c
gapisback: The condition you described is handled by the "hacky-fix" I implemented around this block:

```
 565             if (NumActiveClients <= 0) {
 566                 printf("Server: ThreadID=%" PRIu64 " exiting.\n", tid);
 567
 568                 if (NumActiveClients == 0) {
 569                     req.mtype = REQ_MT_EXIT;
 570
 571                     // Inform last remaining server-thread that this is
 572                     // a final-exit from a server-thread, and not from a real
 573                     // client.
 574                     req.client_idx = -1;
 575                     if (msgsnd(serverId, &req, REQ_MSG_SIZE, 0) == -1) {
 576                         errExit("thread-msgsnd-exit");
 577                     }
 578                 }
```

The problem situation you describe should only occur when the last client exits the system; i.e. `NumActiveClients` transitions from 1 to 0.

- If thread-1 checks the variable and it's non-zero, the thread will end up inside `msgLen = msgrcv` on L515.
- When thread-2 detects on L568 above that the counter has gone to 0, it will send an `REQ_MT_EXIT` message on L575 (which is intended for the other thread-1 waiting for all clients to exit).
- Thread-1 will receive this message and will exit normally.

I have verified this thru test-execution and it seems to work -- at least, works sufficiently well-enough to get the client/server programs to exit cleanly so that perf-tests can be run.

use-cases/client-server-msgs-perf/svmsg_file_server.c
gapisback: We **do need** to invoke the `msgsnd()` to free-up the last remaining thread which is waiting for all clients to exit.

When that call to `msgsnd()` on line above fails, we do do a hard-exit(), so your suggestion is being done already.

use-cases/client-server-msgs-perf/svmsg_file_server.c
gregthelaw: Ah, I get it now! Yes, makes sense.

use-cases/client-server-msgs-perf/svmsg_file_server.c
gregthelaw: Ah yes, I see now. Understood.

use-cases/client-server-msgs-perf/svmsg_file_server.c
gapisback: `__sync_fetch_and_add()` was the earlier interface, but more recently, standard 'C' has also supported use of `Atomic` variables that allow for this kind of `++ ` or `--` avoiding races.

I did some research in this before choosing this approach:

See [this stackoverflow post](https://stackoverflow.com/questions/59662263/making-data-reads-writes-atomic-in-c11-gcc-using-stdatomic-h).

From [cpp-reference on atomic](https://en.cppreference.com/w/c/language/atomic), I see this:

```
The header <stdatomic.h> defines [37 convenience type aliases](https://en.cppreference.com/w/c/thread#Atomic_operations), from [atomic_bool](https://en.cppreference.com/w/c/thread#Atomic_operations) to [atomic_uintmax_t](https://en.cppreference.com/w/c/thread#Atomic_operations), which simplify the use of this keyword with built-in and library types.

_Atomic const int * p1;  // p is a pointer to an atomic const int
const [atomic_int](http://en.cppreference.com/w/c/atomic) * p2;   // same
const _Atomic(int) * p3; // same
```

From [C thread-concurrency pages](https://en.cppreference.com/w/c/thread), I see this:

```
Typedef name 	Full type name
atomic_bool(C11) 	_Atomic _Bool
atomic_char(C11) 	_Atomic char
atomic_schar(C11) 	_Atomic signed char
atomic_uchar(C11) 	_Atomic unsigned char
atomic_short(C11) 	_Atomic short
atomic_ushort(C11) 	_Atomic unsigned short
atomic_int(C11) 	_Atomic int
```

I think for our limited use of 64-bit counters, this use of `_Atomic` should be fine.

Comments welcome.

gapisback:
@gregthelaw - The bulk of this code and test / scripts work is running continually in perf-tests I am exercising.

I am enhancing the perf-report stuff to deal with multiple iterations, etc., all of which will come in a separate commit.

If you can clear the core changes in the 2 main commits, I'd like to integrate that so I don't have to keep working with under-dev code in my dev branches.

gregthelaw:
@gapisback any thoughts on https://github.com/undoio/l3/pull/75#discussion_r1639671179 ?

gregthelaw (approved):

PR #80, opened 2024-06-14
Log fast single threaded improves

gregthelaw:
An optimisation for single-threaded; gets the overhead of a `l3_log_fast()` down to 2ns on my laptop.

Also fixes the bodge I made for accessing the tid in TLS in l3.S.


gapisback (approved):
Changes seem reasonable ... disclaimer: I don't follow the assembly code and take this upon faith.

PR #85, opened 2024-06-16
[Perf] Enhance test.sh and perf_report.py to support multiple iterations.

gapisback:
In previous performance test-runs, we were able to only run one iteration of the client-server program for each run-type, test-parameter. We were seeing some fluctuations and unexpected results in the comparison report.

This commit enhances the testing framework to:
 a) Run 1 or more iterations of the client/server program
 b) Extend the Python perf-comparison report generator to work
    with another source of input metrics-line(s).
 c) Comparison report generated using median value of the
    relevant metric across the n-iterations.

Note: (b) was needed as runs on AWS were aborted midway and had to be partially re-run. We add Python parsing logic to extract out and massage metrics from "Summary: " output lines, that the server program emits to `stdout`.

- To run multiple iterations, use env-var L3_PERF_TEST_NUM_ITERS=<n>

Workflow to run the enhanced perf_report.py:

  - Run test.sh with diff server-threads and #iterations.

Example:
 L3_PERF_TEST_NUM_ITERS=5 L3_PERF_SERVER_NUM_THREADS="1 2 4 8" ./test.sh --clock-default $((1000 * 1000))

  - Redirect output to a /tmp/file: > /tmp/perf.out 2>&1

  - grep 'Summary:' /tmp/perf.out > /tmp/perf.summary.out

  - Feed the summary file to the Python tool using:

    ./scripts/perf_report.py --summary-file /tmp/perf.summary.out

- tests/pytests/perf_report_test.py:

  - Add unit-test cases for new parsing methods added

### DIFF
--- a/scripts/perf_report.py
+++ b/scripts/perf_report.py
@@ -9,9 +9,41 @@ Copyright (c) 2024
 """
 import sys
 import argparse
+import statistics
 from collections import OrderedDict
 from collections import namedtuple
 from prettytable import PrettyTable
+
+# ##########################################################
+# Layout of a summary row from the output file:
+# ##########################################################
+# pylint: disable-next=line-too-long
+# Summary: For 32 clients, Baseline - No logging, num_threads=1, num_ops=32000000 (32 Million) ops, Elapsed time=105613818950 (~105.61 Billion) ns, Avg. Elapsed real time=3300 ns/msg, Server throughput=302990 (~302.99 K) ops/sec, Client throughput=10035 (~10.03 K) ops/sec, Avg. ops per thread=32000000 (32 Million) ops/thread
+
+# Define indexes for each field in above row-format
+FLD_NUM_CLIENTS_IDX         = 0
+FLD_RUN_TYPE_IDX            = 1
+FLD_NUM_THREADS_IDX         = 2
+FLD_NUM_OPS_IDX             = 3
+FLD_ELAPSED_TIME_IDX        = 4
+FLD_AVG_ELAPSED_TIME_IDX    = 5
+FLD_SERVER_THROUGHPUT_IDX   = 6
+FLD_CLIENT_THROUGHPUT_IDX   = 7
+FLD_AVG_OPS_PER_THREAD_IDX  = 8
+
+# -----------------------------------------------------------------------------
+# Define a named tuple class for metrics extracted out from a summary line.
+# pylint: disable-next=line-too-long
+MetricsTuple = namedtuple('MetricsTuple', 'svr_value cli_value svr_units_val svr_units_str cli_units_val cli_units_str thread_str')
+
+# Define a named tuple class for aggregated metrics for a given metric value.
+# niters is the # of raw-data-rows that went into computing the aggregates.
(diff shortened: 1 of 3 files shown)


scripts/perf_report.py
```diff
@@ -9,10 +9,42 @@
...
+# Define indexes for each field in above row-format
+FLD_NUM_CLIENTS_IDX         = 0
+FLD_RUN_TYPE_IDX            = 1
+FLD_NUM_THREADS_IDX         = 2
+FLD_NUM_OPS_IDX             = 3
+FLD_ELAPSED_TIME_IDX        = 4
+FLD_AVG_ELAPSED_TIME_IDX    = 5
+FLD_SERVER_THROUGHPUT_IDX   = 6
+FLD_CLIENT_THROUGHPUT_IDX   = 7
+FLD_AVG_OPS_PER_THREAD_IDX  = 8
```
gapisback, line 32: New parsing interfaces to pluck metrics fields from comma-separated `Summary: ` line in stdout.

scripts/perf_report.py
```diff
@@ -277,6 +316,412 @@ def gen_perf_report(heading:list, metrics:dict, base_svr_value:int, base_cli_val
...
+    list_of_runtypes = gen_runtypes_list(flines)
+    # DEBUG: print(list_of_runtypes)
+
+    base_svr_value = 0
+    base_cli_value = 0
+
+    # Run through diff server-thread parameters, and generate report
+    # We make multiple passes extracing metrics for each combination
+    # of #-of-server-threads and run-types.
+    for threads in list_of_nthreads:
```
gapisback, line 386: This new approach simplifies the post-processing.

We make multiple passes on the output lines, looking for lines of interest to consolidate and to generate aggregate metrics from.

test.sh
```diff
@@ -63,9 +63,12 @@ SvrPerfOutfileArg=
 # execution, we can now specify a list of server-threads to execute,
 # using env-var as:
 #
-#   SERVER_NUM_THREADS="1 2 4 8" ./test.sh run-all-client-server-perf-tests
+#   L3_PERF_SERVER_NUM_THREADS="1 2 4 8" ./test.sh run-all-client-server-perf-tests
```
gapisback, line 66: Slight change in env-var: Use `L3_PERF` prefix, to namespace the env-vars used by this script.

scripts/perf_report.py
```diff
@@ -52,7 +89,7 @@ def do_main(args:list):
     baseline_by_threads = OrderedDict()
 
     # Key: Run-type 'Baseline - No logging'; Value: [ <metrics> ]
-    metrics_by_run = OrderedDict()
+    # metrics_by_run = OrderedDict()
```
gregthelaw, line 92: Is this deliberately committed?

gregthelaw (approved):
Disclaimer: I've only skimmed over this, but it all looks pretty useful / non-controversial.

scripts/perf_report.py
gapisback: Don't recall. Cruft. Cleaned it up.

PR #90, opened 2024-06-19
More fixes to use of TLS from assembly code.

gregthelaw:
My previous attempt to fix usage of TLS wasn't right because child threads would not have their TLS thread initialised. This PR fixes that by zero-initialising the TLS thread, and checking it from every call.

### DIFF
--- a/src/l3.c
+++ b/src/l3.c
@@ -184,7 +184,7 @@ getBaseAddress() {
 }
 #endif  // __APPLE__
 
-L3_THREAD_LOCAL pid_t l3_my_tid;
+L3_THREAD_LOCAL pid_t l3_my_tid = 0;
 
 /**
  * ****************************************************************************
@@ -278,8 +278,6 @@ l3_init(const char *path)
             return -1;
         }
     }
-
-    l3_my_tid = L3_GET_TID();
 
     l3_log = (L3_LOG *) mmap(NULL, sizeof(*l3_log), PROT_READ|PROT_WRITE,
                              MAP_SHARED, fd, 0);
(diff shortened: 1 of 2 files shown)


src/l3.c
```diff
@@ -398,6 +396,7 @@ l3_log_mmap(const char *msg, const uint64_t arg1, const uint64_t arg2,
     int idx = __libc_single_threaded ? l3_log->idx++
                                      : __sync_fetch_and_add(&l3_log->idx, 1);
 #endif  // __APPLE__
+    if (!l3_my_tid) l3_my_tid = L3_GET_TID();
```
gapisback, line 399: Can we please conform to this coding style with enclosing `{}` 

```
399     if (!l3_my_tid) {
400         l3_my_tid = L3_GET_TID();
401     }
```

gapisback (approved):
I took a quick look. Minor change in .c code.

I don't fully understand the bug or the fix in `.s` code. The issue seems to be that some thread-specific `tid` is not coming out right, yes?

If you can take a few minutes and archive the following output:

- A log message from a child-thread showing the wrong / 0 TID before the fix
- A log message from a child-thread showing the wrong / 0 TID with the fix

It will help visually "confirm" that the fix is doing what it claims to do.

Is it easy / difficult to generate this kind of before / after outputs?

gregthelaw:
Prior to this fixup the tid would be the pid when using `l3_fast_log()`. I was able to observe this only from https://github.com/undoio/l3/pull/91. I can't see where the logs get written with `test.sh` but I expect it to be observable there for the the multi-threaded cases.

PR #91, opened 2024-06-19
Add some multi-threaded micro-benchmarks.

gregthelaw:
Spawn 10 threads, each does a million logs.
We have two: one for L3 and one for spdlog.

For this test, L3 is over 20x faster than spdlog.

```
$ time ./spdlog.exe 

real    0m18.050s
user    0m29.033s
sys     1m17.579s
$
$
$ time ./l3.exe 

real    0m0.814s
user    0m6.623s
sys     0m0.028s
$
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -240,12 +240,18 @@ jobs:
         BUILD_MODE=${{ matrix.build_type }} ./test.sh test-build-and-run-csperf-l3_loc_eq_2
 
     #! -------------------------------------------------------------------------
+    # RUN_ON_CI=1 to trigger right selection of spdlog library path on CI-jobs
+    # that run on Mac/OSX machines.
     - name: test-build-and-run-spdlog-sample
       run: RUN_ON_CI=1 ./test.sh test-build-and-run-spdlog-sample
 
     #! -------------------------------------------------------------------------
     - name: test-build-and-run-source-location-cpp20-sample
       run: ./test.sh test-build-and-run-source-location-cpp20-sample
+
+    #! -------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


gapisback:
@gregthelaw - Is this ready to go? This kind of u-perf benchmarking will be interesting to the CS-audience I am giving a talk to this Fri.

I found a few things missing. There is at least an arg-parsing error. You need to apply this diff to get things to run w/o any args:

```
+    int nthreads = argc >=1 ? 10 : atoi(argv[1]);
```

Also I had to change the Makefile to add `-l fmt -lpthread` to the spdlog build step. Did it work for you w/o this?

I have several ideas on how-to productize this with better arg-parsing and a target `mt_ubench_all` target in top-level `Makefile`.

Before I get on and tweak this change-set, wanted to know how far from ready are you on this work?

gregthelaw:
Hi. Thanks for spotting the stupid mistake with args parsing.

The Makefile worked for me on Ubuntu 22.04.04 as it seems on that distro you don't need the -lpthread but yes, older distros will need that, so I fixed that here.

Should be ready for review now.

gapisback:
@gregthelaw - To be consistent with other Makefile targets, I am trying to rework the interfaces to be something on the lines of:

`make all-mt-ubench && make run-mt-ubench`

I think it will be better if we have these targets from top-level `Makefile`.

I am reworking the Makefile to support this. (But it's happening in spurts and during breaks.)

Give me some time to get back to you with a delta-commit offering these changes.

gapisback (approved):
@gregthelaw - I took a quick look. I have updated this change-set in an upcoming PR #92 to streamline the `make` targets into top-level `Makefile`.

You can check this in if you wish. Then, my productization changes will come soon after.

PR #94, opened 2024-08-07
Separate test code into its own directory.

gregthelaw:
This is to make it as easy as possible for someone coming fresh to l3 to just clone the repo and get going.


tests/l3_test.c
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/**
+ * ****************************************************************************
+ * l3_test_log_init() - Initialize L3-logging sub-system, selecting the type of
+ * logging to be performed.
+ * ****************************************************************************
+ */
+int
+l3_test_log_init(const l3_test_log_t logtype, const char *path)
```
gapisback, line 90: I don't follow why you moved this to this file and renamed it as `l3_test_...` . That gives it a flavour that this function is either testing `log_int()` [ which it is not ]. Or, appears that this function is supposed to be under the test-cases.

But, I thought by adding fprintf(), write() and other interfaces to the Client-Server logging programs, we had sort-of agreed that maybe L3 could be extended to allow those logging interfaces, too. Yes?

If so, `l3_log_init()` was meant to the be the initialization interface.

If your objective now is to 'release' L3 only with in-memory mmap() logging, and leave fprintf() / write() purely for benchmarking purposes -- that's fine too.

Again, I am getting out-of-touch with this package and don't have too much bandwidth to "argue" and resolve this with you. So, will let you decide whichever way to go.

tests/l3_test.c
```diff
@@ -0,0 +1,185 @@
...
+}
+
+/**
+ * ****************************************************************************
+ * l3_test_log_deinit() - De-initialize L3-logging sub-system, selecting the type
+ * of logging that was being performed. Finalize the file and close its handle.
+ * ****************************************************************************
+ */
+int
+l3_test_log_deinit(const l3_test_log_t logtype)
```
gapisback, line 121: Same above comments apply to this renamed `deinit()` interface as well.

tests/test.sh
```diff
@@ -598,7 +598,7 @@ function test-build-and-run-csperf-l3-log()
         # No LOC-encoding is in-effect, so no need for the --loc-binary argument.
         ./l3_dump.py                                                                \
                 --log-file /tmp/l3.c-server-test.dat                                \
-                --binary "./build/${Build_mode}/bin/use-cases/svmsg_file_server"    \
+                --binary "tests/build/${Build_mode}/bin/use-cases/svmsg_file_server"    \
```
gapisback, line 601: Would be good to consistently stick with the convention I've tried to follow; i.e. `./tests/build/...`

tests/test.sh
```diff
@@ -646,7 +646,7 @@ function test-build-and-run-csperf-fast-log-no-LOC()
...
```
gapisback, line 649: Same; use `./tests/build/...`

tests/test.sh
```diff
@@ -756,8 +756,8 @@ function test-build-and-run-csperf-l3_loc_eq_1()
         set -x
         L3_LOC_ENABLED=1 ./l3_dump.py                                   \
                             --log-file /tmp/l3.c-server-test.dat        \
-                            --binary "./build/${Build_mode}/bin/use-cases/svmsg_file_server"    \
-                            --loc-binary "./build/${Build_mode}/bin/use-cases/client-server-msgs-perf_loc" \
+                            --binary "tests/build/${Build_mode}/bin/use-cases/svmsg_file_server"    \
+                            --loc-binary "tests/build/${Build_mode}/bin/use-cases/client-server-msgs-perf_loc" \
```
gapisback, line 760: Same for both ... use: `./tests/build/...`

tests/test.sh
```diff
@@ -1121,9 +1121,9 @@ function test-pytests()
 # #############################################################################
 function test-build-and-run-spdlog-sample()
 {
-    make clean && CC=g++ LD=g++ BUILD_VERBOSE=1 make spdlog-cpp-program
+    make -C tests clean && CC=g++ LD=g++ BUILD_VERBOSE=1 make -C tests spdlog-cpp-program
 
-    local test_prog="./build/${Build_mode}/bin/use-cases/spdlog-Cpp-program"
+    local test_prog="tests/build/${Build_mode}/bin/use-cases/spdlog-Cpp-program"
```
gapisback, line 1126: Multiple such places where it would be nice[r] to replace: `./tests/build/.../`

Çan you please fix all such instances?

gapisback (approved):
Made one quick pass on the diffs. Cloned the repo on my Mac and gave the tests a quick-run. Seem to work ok.

I 've left behind some cleanup comments. Mainly, it will be helpful to add a `test.md` in Docs/ folder.

Please resolve all comments as best possible per your choices and kindly move ahead to merge. Unless you have a strong reason for re-review, I probably won't have much more time, till next weekend.

tests/Makefile
```diff
@@ -134,29 +134,29 @@ CXX_X_FLAG  := c++
 # ###################################################################
 #
 L3PACKAGE   := l3
-SRCDIR      := src
-INCDIR      := include
```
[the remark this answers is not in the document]
gregthelaw: Yes, I think it's better this way. I want someone who is unfamiliar with the project to be able to clone the repo and see immediately that there are four files in the top level dir that they care about - a .c file, a .h file, a .S file (if they want it) and l3_dump.py. That's all most people will care about, so it makes it easer to know what's important if everything that most people won't care about is squirrelled away in directories.

tests/test.sh
```diff
@@ -221,60 +221,60 @@ function test-pylint-check()
 # #############################################################################
 function test-make-help()
 {
-    make help
+    make -C tests help
```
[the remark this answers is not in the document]
gregthelaw: Good idea - done

tests/l3_test.c
gregthelaw: Apologies for giving the impression that I'd agreed with the approach, I hadn't intended to.

I think for the library to get traction, we should make it as simple and as obvious as possible. If we want some kind of dynamic logging switching layer, I think that should be a separate GH repo. Being able to switch between them as we do for our testing/benchmarking code is super useful _for us_. But for 99% of people it will be a distraction. They just want a simple, fast logging library to use. Every bit of complexity / thing for users to have to have to learn/understand will reduce the amount the library gets used.

I've learnt all this the hard way with UDB. You have to make it soooo simple and straightforward for people to adopt. People just will not read documentation.

gregthelaw: I'm sure this stuff can all be made a lot better / cleaner, but the urgency is to get all our testing/benchmarking code "out of the way" so that users don't need to pollute their brainspace with it. Then I'll publish the video I have on this, and do more pushing on the socials etc (using the Undo marketing dept to do this) and hopefully we can get a little bit of traction.

tests/l3_test.c
gregthelaw: As above.

tests/test.sh
gregthelaw: You mean start with `./`? Sure, done.

tests/test.sh
gregthelaw: done

tests/test.sh
gregthelaw: done

PR #95, opened 2024-08-12
L3 log fast via weak symbols

gregthelaw:
We make the C implementation of `l3_log()` use a weak symbol, and rename the assembly version of the routine to `l3_log()`. This way, if the assembly version is present, we'll pick that up, and the caller doesn't need to worry about which version to use.

(I'm still not convinced the assembly version is sufficiently faster than the C version to make all this worthwhile, but ...)


l3_dump.py
```diff
@@ -500,7 +500,11 @@ def do_main(args:list, return_logentry_lists:bool = False):
             # print(f"{msgptr=:x}, {fibase=:x}, {rodata_offs=:x}, {offs=}")
 
             # Generate C-style sprintf() output on message-string.
-            msg_text = do_c_print(strings[offs], arg1, arg2)
+            try:
+                msg_text = do_c_print(strings[offs], arg1, arg2)
+            except KeyError:
```
gapisback, line 505: This change _seems_ reasonable, except in the callee, I see this comment block:

```
382     If there is any exception in printing, fall-back to just dumping the
383     input arguments, without any modifications.
```

This fall-back behavior I had coded / intended for now conflicts with the new error raising you are implementing.

My thinking in coding the fall-back behavior was that in the user's program, we should expect a properly formatted print format-string. So, what's the likelihood of getting an error?

I don't know how or where KeyError exception will be raised.

I am not sure how to sort out what you are trying to do v/s what I had coded earlier.

---
I will guide you to the pytests in `pytests/l3_dump_parse_test.py`. 

See test_fmtstr_replace() and test_fmtstr_replace_and_print().

From code coverage point of view, can you investigate if it's possible to add a new pytest invoking `do_c_print()`, i.e. add `test_do_c_print()`, ... and see if there is some sense in testing out by invoking an exception raising situation.

I don't have time to sort out the pytest angle. It's just an idea for you to follow-thru as best possible.

README.md
```diff
@@ -7,8 +7,7 @@ and usually involves acquiring a lock, so it becomes a kind of
 synchronisation point.
 
 The Lightweight Logging Library ([l3](./include/l3.h)) has a fast but limited `printf` API:
-- `l3_log_simple()`: Takes about 10ns (Intel Core i7-1365U)
-- `l3_log_fast()`: An even faster, but more limited interface, which takes about 7ns.
+- `l3_log()`: Takes about 2ns (Intel Core i7-1365U)
```
gapisback, line 10: ? Is the updated comment about `2ns` accurate?

In old comment, C-fn was cited to take 10ns and assembly to take ~7ns.

Where did you get this 2ns from?

Kindly investigate and correct this number. Otherwise, future readers will be mislead, no ?

tests/perf/mt_ubench_l3.cpp
```diff
@@ -46,7 +46,7 @@ int main(int argc, char** argv) {
             assert(e == 0 || e == PTHREAD_BARRIER_SERIAL_THREAD);
 #endif
             for (int j = 0; j < L3_PERF_UBM_NMSGS; j++) {
-                l3_log_fast("Hello, world! %d %d", 0, j);
+                l3_log("Hello, world! %d %d", 0, j);
```
gapisback, line 49: This is nice! As a user, I don't need to remember fast or slow, and the .s version is just picked up if it exists for the current platform. Hope this simplifies adoption in months to come ...

gapisback (approved):
Took a fresh look and put down some minor comments. Nothing is a blocker. Kindly resolve as you feel best to move this along.

In [this commit](https://github.com/undoio/l3/pull/95/commits/6bac966aa5b7a897c9d46eda4f831ff3692979aa), there was some refactoring of perf-metrics code. I didn't review this very closely. Am assuming that you did re-run the u-perf and other perf-type tests, and confirmed that the right summary metrics are being printed, as needed.

l3_dump.py
gregthelaw: Hmm, I see what you're saying, I think you're right, this should never be user visible. I'll drop this commit from the PR.

README.md
gregthelaw: Yes, this is correct. An earlier PR improved the performance. At least for single-threaded programs. I'll expand the text to be clearer.

tests/perf/mt_ubench_l3.cpp
gregthelaw: thanks!

